PR #3, opened 2019-02-13
calling Alloy from Haskell

marcellussiegburg:
Make it so that in `random-cd2alloy` there is a call to the Alloy Analyzer at the end (in a suitable form, maybe a Java executable wrapping Alloy), on the concatenation of files `output.part1` ... `output.part5`, and all (or a certain number of) instances in text form are generated.


jvoigtlaender:
This is now to happen in `random-stuff` instead of `random-cd2alloy`.

jvoigtlaender:
And the relevant file is directly the complete "output.als", not in parts anymore.

jvoigtlaender:
Do it by changing the `giveMeInstances` stub I just added and committed.

random-stuff.hs
```diff
@@ -1,38 +1,54 @@
+{-# LANGUAGE NamedFieldPuns #-}
+
```
jvoigtlaender: what for?

random-stuff.hs
```diff
@@ -1,38 +1,54 @@
...
+  drawCdFromSyntax syntax (output config) Pdf
+  unless (anyRedEdge syntax) $ do
     time <- getZonedTime
     let (part1, part2, part3, part4, part5) = transform syntax "" (show time)
-    let out = output ++ ".als"
+    let out = output config ++ ".als"
     writeFile out (part1 ++ part2 ++ part3 ++ part4 ++ part5)
-    putStrLn ("More output written to " ++ out)
-    instances <- giveMeInstances
+    putStrLn ("More output written to " ++ output config)
```
jvoigtlaender: Isn’t out the correct file name?

random-stuff.hs
marcellussiegburg: I guess I missed that on manual merging without history. It is not required.

jvoigtlaender (commented):
Thanks

PR #4, opened 2019-02-13
something still missing on Java side

marcellussiegburg:
Maybe need to add the class file?

I get:
`Fehler: Hauptklasse RunAlloy.java konnte nicht gefunden oder geladen werden`


marcellussiegburg:
I have no class file and it is working. I can not tell what the issue is.

jvoigtlaender:
Maybe `javac` is silently called on your system if you call `java` with only a `.java` file present?

I don't even have `javac` on my system.

Can you make a `.class` file and commit it?

marcellussiegburg:
Maybe it is required to tell java to look within the same folder?
i.e. instead of
```Haskell
  let callAlloy = proc "java" ["-cp", "Alloy-5.0.0.1.jar", "RunAlloy.java", output c ++ ".als", show $ maxInstances c]
```
either
```Haskell
  let callAlloy = proc "java" ["-cp", "Alloy-5.0.0.1.jar:.", "RunAlloy.java", output c ++ ".als", show $ maxInstances c]
```
or

```Haskell
  let callAlloy = proc "java" ["-cp", "Alloy-5.0.0.1.jar;.", "RunAlloy.java", output c ++ ".als", show $ maxInstances c]
```
?

marcellussiegburg:
Okay, I will create one.

jvoigtlaender (commented):
Okay. Meanwhile also installed a JDK on my system. Will see whether building from source works as well.

PR #5, opened 2019-02-13
avoid creating the .als file?

marcellussiegburg:
I guess instead of manifesting `output.als` in the file system, things could be organised such that the string is directly piped to the `RunAlloy` process, which then reads it from `stdin`?


marcellussiegburg:
Yes, this should be possible.

PR #8, opened 2019-02-14
select path separator at compile time

marcellussiegburg:


jvoigtlaender:
This doesn't work. The value of `System.FilePath.Posix.searchPathSeparator` on my system is `':'`.

jvoigtlaender:
Shouldn't you use `System.FilePath.searchPathSeparator` instead?

marcellussiegburg:
Yes I guess so, but in that case the haddock
http://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Posix.html#v:searchPathSeparator
is misleading.

jvoigtlaender:
Though, at the top of that page ...

marcellussiegburg:
Can you try this on Windows?

jvoigtlaender:
Works

PR #14, opened 2019-02-21
Fix overflow issues in alloy

marcellussiegburg:
- prevent overflow on int in alloy by changing alloy options
 - scale integer range based on number of classes and maximum number of objects


jvoigtlaender:
Oh, well, did you test this? :smile:

Making that change in `Transform.hs` has no effect, since in `random-task` the `part5` portions are thrown away.

marcellussiegburg:
Good point, I can not tested it, however I did not test the complete process.

I will write a fix.

PR #20, opened 2019-02-22
Enable full exercise generation

### LINKED ISSUE #15: implement generation of a full exercise task
... replacing the current `random-task` implementation.

Here is how I envision the process, given a `ClassConfig`:

1. Generate a random class diagram `cd1`. It is fine if it has `anyRedEdge`.
2. From `cd1`, generate a `cd2` via random mutation, still within the given `ClassConfig`. It must not have `anyRedEdge`, so if necessary repeat 2. (Or if no valid mutations exist, go back to 1.)
3. Again from `cd1`, generate a `cd3` via random mutation, still within the given `ClassConfig`. It must not have `anyRedEdge`, so if necessary repeat 3. (No need to check whether `cd2` and `cd3` are different. If they happen to be equal, that will be flagged later on, anyway.)
4. Generate instances for `cd2 and not cd3`. Shuffle them randomly.
5. Generate instances for `cd3 and not cd2`. Shuffle them randomly.
6. Generate instances for `cd2 and cd3`. Shuffle them randomly.
7. We are aiming to take `5` instances overall from the fronts of the lists produced in 4., 5., 6., subject to the constraint that at most `3` are taken from the same list. If that is not possible, go back to 1. If it is possible, make a random such selection.
8. Create pictures for:
    - `cd2`
    - `cd3`
    - the `5` selected instances, named in such a way that we know for each which of the lists 4., 5., 6. it came from

marcellussiegburg:
- close #15

### DIFF
--- a/random-task.hs
+++ b/random-task.hs
@@ -45,9 +45,9 @@ getRandomTask config output searchSpace maxInstances = do
     instances3not2 <- getInstancesOfMerged parts2 parts3 cd3not2
     instances2and3 <- getInstancesOfMerged parts2 parts3 cd2and3
     let takes = [ (take x, take y, take z)
-                | x <- [1 .. min 3 (length instances2not3)]
-                , y <- [1 .. min 3 (length instances3not2)]
-                , z <- [1 .. min 3 (length instances2and3)]
+                | x <- [0 .. min 3 (length instances2not3)]
+                , y <- [0 .. min 3 (length instances3not2)]
+                , z <- [0 .. min 3 (length instances2and3)]
                 , 5 == x + y + z ]
     continueIf (not $ null takes) $ do
       (take2not3, take3not2, take2and3) <- head <$> shuffleM takes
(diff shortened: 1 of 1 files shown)


random-task.hs
```diff
@@ -24,36 +23,66 @@ main = do
...
+        parts3 = transform cd3 "3" ""
+        cd2not3 = createRunCommand "cd2 and (not cd3)" (length names) 5
+        cd3not2 = createRunCommand "cd3 and (not cd2)" (length names) 5
+        cd2and3 = createRunCommand "cd2 and cd3" (length names) 5
+    instances2not3 <- getInstancesOfMerged parts2 parts3 cd2not3
+    instances3not2 <- getInstancesOfMerged parts2 parts3 cd3not2
+    instances2and3 <- getInstancesOfMerged parts2 parts3 cd2and3
+    let takes = [ (take x, take y, take z)
+                | x <- [1 .. min 3 (length instances2not3)]
+                , y <- [1 .. min 3 (length instances3not2)]
```
jvoigtlaender: Why `1 ..`? I didn't rule out `0`.

jvoigtlaender:
Great! Next stop, after addressing #19, an Autotool exercise task using the same process.

jvoigtlaender:
:smile:

PR #24, opened 2019-03-10
Split random task generation

marcellussiegburg:


jvoigtlaender:
I guess the next refactoring step is to move `getRandomTask` and the functions it depends on into a separate module?

jvoigtlaender:
Actually, I would have expected a refactoring where `shuffleM` is not used with `IO` at all, to make sure its randomness can be set from the program without relying on `IO`.

Split `getRandomInstances` further, into one function that has no `IO` type, but has all the `shuffleM` calls, and another function that has an `IO` type (and the `Alloy.getInstances` calls)?

PR #36, opened 2019-11-07
Match edges between CD and OD

marcellussiegburg:
- Adds the ability to create tasks for matching edges of a CD to associations of an OD


jvoigtlaender:
I guess I should have dealt with https://github.com/jvoigtlaender/alloy-cd-od/pull/35 at some point, or what is the status of that PR?

The two probably conflict. Should I still merge this one here? Guess I would want to start playing with it a bit.

marcellussiegburg:
We can keep #35 around. Yes it would have probably conflicts because the drawing mechanism was required to change for these changes here. This pull request here has a higher priority. We can switch to the other drawing mechanism later. (I.e. handle #35 at some point later)

jvoigtlaender:
Should I squash the commits here? Maybe better not. But rebase from the GitHub interface does not work here, probably too many things at odds. Do you want to rebase on the command line?

marcellussiegburg:
No, better not. I will do a rebase and force push it.

Issue #40, opened 2019-11-12
long term: change Alloy modelling with respect to navigation directions of aggregations/compositions


jvoigtlaender:
See https://github.com/jvoigtlaender/alloy-cd-od/issues/38#issuecomment-552336372.

jvoigtlaender:
See also https://github.com/jvoigtlaender/alloy-cd-od/pull/41#pullrequestreview-319264110.

jvoigtlaender:
Also need to take into account consequences for "external" uses of the modelling, specifically the various direction fixups/hacks here:
- https://github.com/fmidue/autotool-to-xmodeler/blob/b761bcd70668d7e50d662a1b0cc997b7052b132f/app/fromAutotool.hs#L73-L74
- https://github.com/fmidue/autotool-to-xmodeler/blob/b761bcd70668d7e50d662a1b0cc997b7052b132f/app/fromAutotool.hs#L83-L86

marcellussiegburg:
What remains to be done here is amendment of `autotool-to-xmodeler` (see https://github.com/jvoigtlaender/modelling-tasks-internal/issues/40#issuecomment-1386768285). Is there a specific test case that can be used to guide amendments in that foreign repository?

jvoigtlaender:
That code hasn't been used for a while. I think it is okay to simply open an issue there that points out things would have to be changed due to major changes in `modelling-tasks`. Same applies to [fmidue/xmodeler-tooling](https://github.com/fmidue/xmodeler-tooling).

jvoigtlaender:
(It's not just the directedness that has changed and would have consequences in those two repos. Also other aspects of the Alloy modelling, such as the different embedding strategy etc.)

marcellussiegburg:
The issues are now tracked in fmidue/autotool-to-xmodeler#1 and fmidue/xmodeler-tooling#9

PR #41, opened 2019-11-19
Create Alloy module to generate class diagrams

marcellussiegburg:
- the generation includes the generation of changes on the
   generated class diagram
 - the generation may be configured to
   - generate wrong class diagrams
   - generate fixes for wrong class diagrams
   - generate changes only on existing class diagrams


generate.als
```diff
@@ -0,0 +1,335 @@
...
+  validToLimitsAssoc [a]
+}
+
+pred validLimitsComposition [a : Assoc] {
+  (a.toLower = Zero or a.toLower = One) and a.toUpper = One
+}
+
+sig Inheritance extends Relationship {}
+
+abstract sig Assoc extends Relationship {
```
jvoigtlaender: `Assoc`s don't need names, because that will be handled (i.e., names be assigned from some reservoir) on the Haskell side?

marcellussiegburg: Yes, this will be handled on the Haskell side.

generate.als
```diff
@@ -0,0 +1,335 @@
...
+  a.toLower = Zero implies a.toUpper != Zero
+}
+
+pred validLimitsAssoc [a : Assoc] {
+  validFromLimitsAssoc [a]
+  validToLimitsAssoc [a]
+}
+
+pred validLimitsComposition [a : Assoc] {
+  (a.toLower = Zero or a.toLower = One) and a.toUpper = One
```
jvoigtlaender: So the direction of compositions (also aggregations, I assume) is now according to https://github.com/jvoigtlaender/alloy-cd-od/issues/40, i.e., reversed over how the Haskell code has treated them so far? That means that when parsing the Alloy instances into Haskell structures for further processing via the existing machinery there, some caution is needed.

generate.als
```diff
@@ -0,0 +1,335 @@
...
+  fromUpper : one Limit,
+  toLower : one Limit,
+  toUpper : one Limit
+}
+
+sig Aggregation extends Assoc {}
+sig Association extends Assoc {}
+sig Composition extends Assoc {}
+
+pred noSelfRelationship [c : Relationship] {
```
jvoigtlaender: I guess `r` would now be a better variable name here and in a few other places.

generate.als
```diff
@@ -0,0 +1,335 @@
...
+  c.from != c.to
+}
```
jvoigtlaender: I'm in favour of predicate names (and negated semantics, of course) `selfRelationship`, `doubleRelationship`, ... `doubleInheritance`.

generate.als
```diff
@@ -0,0 +1,335 @@
...
+
+pred noDoubleRelationship [c, c' : Relationship] {
+  c != c' implies c.from = c'.from implies c.to != c'.to
+}
+
+pred noReverseRelationship [c, c' : Relationship] {
+  c != c' implies c.to = c'.from implies c.from != c'.to
+}
+
+pred noDoubleInheritance [i, i' : Inheritance] {
```
jvoigtlaender: I'd call this "multiple inheritance", not "double inheritance".

generate.als
```diff
@@ -0,0 +1,335 @@
...
+  i != i' implies i.to != i'.to
+}
+
+fun relationship (restriction : set Relationship) : Class -> Class {
```
jvoigtlaender: I don't think I like that the argument is introduced with round brackets here, but then later used with square brackets on invocation.

generate.als
```diff
@@ -0,0 +1,335 @@
...
+
+pred markedEdgeCriterion [xFrom, xTo, yFrom, yTo : Class, is : set Inheritance] {
+  let subs = ~(relationship [is]) |
+    (xFrom != yFrom or xTo != yTo)
+      and (yFrom in xFrom.*subs and (yTo in xTo.*subs or xTo in yTo.*subs)
+        or yTo in xTo.*subs and (yFrom in xFrom.*subs or xFrom in yFrom.*subs))
+}
+
+pred isMarkedEdge [r : Relationship, rs : set Relationship, is : set Inheritance] {
+  some r' : rs - Inheritance |
```
jvoigtlaender: I'd rather see this `- Inheritance` at the invocation site of `isMarkedEdge`.

generate.als
```diff
@@ -0,0 +1,335 @@
...
+    markedEdgeCriterion [r.from, r.to, r'.from, r'.to, is]
+    or markedEdgeCriterion [r.to, r.from, r'.from, r'.to, is]
+}
+
+pred noMarkedEdges [rs : set Relationship, is : set Inheritance] {
+  all r : rs - is | not isMarkedEdge [r, rs, is]
```
jvoigtlaender: Similar comment as for `isMarkedEdge` above.

generate.als
```diff
@@ -0,0 +1,335 @@
...
+fact nonEmptyInstancesOnly {
+  some Relationship
+}
+
+sig Change {
+  add : lone Relationship,
+  remove : lone Relationship
+}
+
+pred sameDirection [r, r' : Relationship] {
```
jvoigtlaender: Isn't this (easily made) redundant, given the existence of `noDoubleRelationship`?

generate.als
```diff
@@ -0,0 +1,335 @@
...
+pred changedKind [c : Change] {
+  sameDirection [c.add, c.remove]
+  not sameKind [c.add, c.remove]
+  c.add in Assoc and c.remove in Assoc implies sameFromLimits [c.add, c.remove]
+  c.add in Assoc and c.remove in Assoc
+    and (validLimitsComposition [c.remove] or not c.add in Composition)
+    implies sameLimits [c.add, c.remove]
+}
+
+pred increasedFromRange [a, a' : Assoc] {
```
jvoigtlaender: I think everything from here to `shiftedRange`, and possibly some of the material/definitions supporting these predicates, would better be moved to a separate module which is about "limits and their manipulation".

It would probably also make sense to add to that separate module, basically as unit tests, some Alloy assertions (not facts) which clarify what the intended semantics of the various predicates here is.

generate.als
```diff
@@ -0,0 +1,335 @@
...
+}
+
+pred shiftedRange [a, a' : Assoc] {
+  shiftedRangeUp [a, a'] iff not shiftedRangeUp [a', a]
+}
+
+pred changedLimit [c : Change] {
+  sameDirection [c.add, c.remove]
+  sameKind [c.add, c.remove]
+  validLimitsAssoc [c.add]
```
jvoigtlaender: What happens here if `c.add` is not an `Assoc`?

generate.als
```diff
@@ -0,0 +1,335 @@
...
+}
+
+pred change [c : Change, rs : set Relationship] {
+  some c.add + c.remove
+  no c.add or not c.add in rs
+  c.remove in rs
+  let c1 = changedLimit [c],
+      c2 = changedKind [c],
+      c3 = flip [c] |
+    one c.add and one c.remove iff c1 implies not c2 and not c3 else c2 iff not c3
```
jvoigtlaender: Couldn't the right-hand side of this `iff` simply be `c1 or c2 or c3`?

generate.als
```diff
@@ -0,0 +1,335 @@
...
+}
+
+fact changesAreUnique {
+  all c, c' : Change | c != c' implies (c.add = c'.add implies c.remove != c'.remove)
```
jvoigtlaender: Couldn't the part after `implies` here simply be `c.add != c'.add or c.remove != c'.remove`?

And wouldn't applying the same rewrite to the next line actually allow to simplify the whole logical formula here?

(Similarly elsewhere, I assume.)

generate.als
```diff
@@ -0,0 +1,335 @@
...
+    and (c.remove = c'.remove implies c.add != c'.add)
+}
+
+fact preventSaneErrors {
```
jvoigtlaender: It's unclear to me what this is for.

generate.als
```diff
@@ -0,0 +1,335 @@
...
+  relationships : set Relationship,
+  wrongAssocs : one Int,
+  wrongCompositions : one Int,
+  selfRelationships : one Int,
+  hasDoubleRelationships : one Boolean,
+  hasReverseRelationships : one Boolean,
+  hasDoubleInheritances : one Boolean,
+  hasInheritanceCycles : one Boolean,
+  hasCompositionCycles : one Boolean,
+  hasMarkedEdges : lone Boolean] {
```
jvoigtlaender: What is the motivation for having this `lone`, rather than `one` as all the other Booleans?

(Also below in `changeOfFirstCD`.)

generate.als
```diff
@@ -0,0 +1,335 @@
...
+    let Assoc' = Assoc - (Change.add - c.add) - c.remove,
+        Composition' = Composition - (Change.add - c.add) - c.remove,
+        Relationship' = Relationship - (Change.add - c.add) - c.remove,
+        Inheritance' = Inheritance - (Change.add - c.add) - c.remove {
+      change[c, Relationship - Change.add]
+      classDiagram [Assoc', Composition', Relationship', Inheritance',
```
jvoigtlaender: Isn't that the wrong argument order, given the definition of `classDiagram`?

generate.als
marcellussiegburg: Yes, it is as #40 demands. This requires some caution during import into existing structures. Perhaps, the procedure well be reverting reverse relationships of ec83376ef215007079e59acd325b8c9db470874.

generate.als
marcellussiegburg: - `sameDirection` may evaluate to *true*, as it is a `Relationship`
- `sameKind` evaluates to *true*, if `c.remove` is an `Inheritance`
- `validLimitsAssoc` evaluates to *false* as `c.add` is an `Inheritance` but the predicate `validLimitsAssoc` demands its arguments to be an `Assoc`

-> thus `changedLimit` evaluates to *false*

generate.als
marcellussiegburg: It could, if we allow changes to be a combination of several modifications. However, if we want to keep the behaviour of our existing library they have to be exclusively applied.

generate.als
marcellussiegburg: This predicate is considered to catch semantically overlapping changes.
This prevents formal correct behaviour for a change, which does not have any different meaning to another change (i.e. an equivalent change).
Concretely, `lower = Zero and upper = Star` is semantically the same as `lower = Star and upper = Star`.

generate.als
marcellussiegburg: The idea was to allow a diagram to having or not having marked Edges without forcing either of them. Of course, this may be achieved in a different way as well (as I did for the example generation of class diagrams à la our alloy-cd-od-paper approach (see pedicate changes)).

generate.als
marcellussiegburg: Does this include quantified predicates like `noDoubleRelationships` should they rather use `some` instead of `all` and be renamed to either `doubleRelationships` or `someDoubleRelationships`?

jvoigtlaender: No, does not include them. And instead of `all` they should then (after negating the ones here) use `no`.

generate.als
jvoigtlaender: Is that really the case? Aren't the predicates `changedLimit`, `changedKind`, `flip` mutually exclusive anyway? Can there, for example, ever be a `c` such that both `changedLimit[c]` and `changedKind[c]` are true?

generate.als
jvoigtlaender: Did you check this? The assumed effect of type-checking on semantics?

The Alloy book says something else. Namely, for this case, that the predicate `validLimitsAssoc` will evaluate to whatever its definition body says, rather than evaluating to false simply because it wanted an `Assoc` but got an `Inheritance`.

generate.als
jvoigtlaender: If there are constraints on the relationship between `lower` and `upper` for each `Assoc`, then these would better be put into a signature constraint directly when `Assoc` is defined.

generate.als
marcellussiegburg: You are right, they exclude each other due to the additional constraints made within each of the predicates.

generate.als
marcellussiegburg: I checked it in the given case but maybe some other constraints interfered. A simple isolated check proved me wrong. Thanks for the advice.

generate.als
marcellussiegburg: `doubleRelationship` contains the constraint that `r != r'`, I will reuse `sameDirection` within `doubleRelationship`.

jvoigtlaender: How about moving the `r != r'` constraints out of `doubleRelationship` (and similar places) to their invocation sites? (Actually, I guess I would inline a lot of these predicates, into `noDoubleRelationships` etc.)

generate.als
marcellussiegburg: Fixed by 73d0a4c

generate.als
marcellussiegburg: Fixed by e0eac05

generate.als
marcellussiegburg: Fixed by 1d8f4cb

generate.als
marcellussiegburg: Fixed by 387cf93

generate.als
marcellussiegburg: Addressed by eeab606, isMarkedEdge was changed to take a second parameter of type `set Assoc`, which is, what makes more sense anyway. The calls were changed accordingly.

generate.als
marcellussiegburg: See comment above

generate.als
marcellussiegburg: I think I would rather not move it out of the predicate it is a double relationship only, if the relations are not equal. I guess it would be possible to inline the predicates. However, keeping the predicates allows for violating the constraint for specific relationships later. I am not sure yet if this will be required or not.

generate.als
marcellussiegburg: Fixed by e9c7203 and fbb81e2

generate.als
marcellussiegburg: Fixed by 90f5130

generate.als
marcellussiegburg: Fixed by 9eb55b9

generate.als
marcellussiegburg: Fixed by ea18422. I checked other places as well. Of course replacing `a implies b` by `not a or b` is always an option. However, I think that in it is easier to read the implication in many cases. Therefore, I have kept the implication in other cases. If you insist on removing some or all implications I will implement it. (Sometimes the formula would become longer, sometimes it does not make any difference, I did not find another place where it would reduce the size, I did not check to deeply for other ways of expressing `... implies ... else ...`)

generate.als
marcellussiegburg: Fixed by fc66fbc.

generate.als
marcellussiegburg: Fixed by 66ed7d5

jvoigtlaender:
Now that we have some assertions, I wonder how/when exactly they will be exercised.

During normal instance generation (creating exercise tasks) they will be ignored? But we can have a separate "testing cycle" where we `check` that the assertions actually hold?

Or do we also want to turn some `check`s on during the generation of concrete tasks? What would happen then if one of the assertions is actually found to be false? A runtime error?

jvoigtlaender:
About this pull request: Should it be merged as is? Or do you want to move the two `.als` files to some subdirectory or so?

PR #46, opened 2019-12-04
Introduce a task for choosing correct changes that fix a broken class diagram.

marcellussiegburg:


cd/generate.als
```diff
@@ -112,6 +112,14 @@ pred changedLimit [c : Change] {
   shiftedRange [c.add, c.remove] iff not changedRange [c.add, c.remove]
 }
 
+pred sameRelationship [r, r' : Relationship] {
+  r != r'
```
jvoigtlaender, line 116: Somehow I'd expect to not see this here (also taking into account how this predicate is used elsewhere).

PR #53, opened 2020-01-14
Use call alloy library for model instance finding.

marcellussiegburg:


src/CD2Alloy/Transform.hs
```diff
@@ -55,7 +55,7 @@ createRunCommand command numClasses maxObjects = unlines
   , "// Run commands"
   , "///////////////////////////////////////////////////"
   , ""
-  , "run { " ++ command ++ " } for " ++ show maxObjects ++ " Obj, " ++ show intSize ++ " Int"
+  , "run { " ++ command ++ " } for " ++ show maxObjects ++ " Obj, " ++ show intSize ++ " FName, " ++ show intSize ++ " Int"
```
jvoigtlaender, line 58: What is the rationale for using `intSize`, which I assume is the required bitwidth of integers overall, as bound for `FName` as well? I'd guess that they are rather unrelated?

marcellussiegburg: Yes, choosing a more accurate value would probably be best there. I guess it would be best to introduce a new parameter for the number of edges. It could also be enough to assume at most a complete graph (i.e. `n * (n -1) / 2`). (Because we do not allow changes of ODs within the Alloy program yet)

jvoigtlaender: I stumbled upon this in the context of XModeler work.

Why is a scope declaration on `FName` even necessary? Isn't it the case that there are always explicit `one sig ... extends FName` declarations, and thus the scope is automatically determined by Alloy via the number of such `one sig`s?

In particular, the above reference to "number of edges" seems out of place, since `FName` is not for representing edges. It is just for representing possible labels on edges.

marcellussiegburg: In fact this scope declaration is not necessary. In turn, it should be removed, it could be defined as `exactly` number of `one sig ... extends FName` delarations, but as `FName` is `abstract` already, as you said (and according to the Alloy-book), Alloy is able to figure it out by itself. Furthermore restrictions on the number of edges have been introduced later on (using `get`, at the moment: https://github.com/jvoigtlaender/modelling-tasks-internal/blob/dec6377b2726f26381e8809c74302c9c27a90e9e/src/Modelling/CdOd/CD2Alloy/Transform.hs#L95-L103)

PR #62, opened 2020-03-10
Update code according to current autotool implementation

### LINKED ISSUE #55: merge 'move-src' branch
That branch is quite long-lived by now. Shouldn't it be merged into 'master'? Or is there some reason for developing it separately?

Context: https://martinfowler.com/bliki/FeatureBranch.html

marcellussiegburg:
- close #55


src/Alloy/CdOd/CD2Alloy/Transform.hs
```diff
@@ -1,65 +1,77 @@
...
-    , "// Produced by Haskell reimplementation of Eclipse plugin transformation"
-    , "// Generated: " ++ time
-    , ""
-    , "module umlp2alloy/CD" ++ index ++ "Module"
-    , ""
-    , $(embedStringFile "od/template.als")
-    ]
- part2 = unlines
-  [ "// Concrete names of fields"
-  , unlines (associationSigs associations) -- Figure 2.1, Rule 3, part 2
```
jvoigtlaender: I don't like that this comment (and other ones in the same file) are disappearing with this commit.

marcellussiegburg: Shall I reintroduce these comments? Which document are they referencing?

jvoigtlaender: Yes. The document is https://git.rwth-aachen.de/monticore/publications-additional-material/blob/master/cd2alloy/CD2AlloyTranslationTR.pdf.

jvoigtlaender: It might be better to replace the references by ones to stuff in [7] from the SACLA paper, i.e., https://www.se-rwth.de/publications/CD2Alloy-A-Translation-of-Class-Diagrams-to-Alloy.pdf, but that might be non-trivial. I am not sure whether all the material is to be found there identically. Which is of course particularly relevant since some of the comments state that some stuff from the original document needed to be fixed. And I simply don't know (since I wasn't aware of the 2017 version of the document at the time of implementing those rules and putting those comments in) whether [7] from the SACLA paper has those things to be fixed as well...

The point is: When I implemented the CD2Alloy transformation in Haskell, I worked from the information provided at https://www.se-rwth.de/materials/cd2alloy/, which does not include a pointer to the much more recent 2017 TR.

marcellussiegburg: I would suggest additionally referencing at least the document name.

jvoigtlaender: Sure.

PR #169, opened 2022-02-02
CI tweaking

jvoigtlaender:
To some extent, these are cosmetics, but not solely. In particular, the `permissions` setting should make the Hlint CI Action work on the public repo again.

I do have some other questions on the main CI workflow file:

Should it be made much simpler by throwing out all cabal-only stuff? So, switching over to using stack for everything?

What's the role of the `test` argument here: https://github.com/jvoigtlaender/modelling-tasks-internal/blob/f90f4d1743afea42b79a48fdba258765be9a074b/.github/workflows/haskell.yml#L89 and here: https://github.com/jvoigtlaender/modelling-tasks-internal/blob/f90f4d1743afea42b79a48fdba258765be9a074b/.github/workflows/haskell.yml#L109 ?

If these commands are for installing dependencies only, there is nothing to test? Or are test suites of the dependencies run as well?

What is the rationale for the `--only-dependencies` argument here: https://github.com/jvoigtlaender/modelling-tasks-internal/blob/f90f4d1743afea42b79a48fdba258765be9a074b/.github/workflows/haskell.yml#L168 ?

It seems the following line is the "parallel" (for Unix instead of Windows) of the above, and it does not have `--only-dependencies`: https://github.com/jvoigtlaender/modelling-tasks-internal/blob/f90f4d1743afea42b79a48fdba258765be9a074b/.github/workflows/haskell.yml#L134

And in the above two lines, would it be useful to add `--coverage` after `test`?

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -11,7 +11,7 @@ on:
       - master
 
 jobs:
-  build:
+  consistency:
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -11,7 +11,7 @@ on:
       - master
 
 jobs:
-  build:
+  build_and_test:
     strategy:
       matrix:
 #        os: [ubuntu-latest, macOS-latest, windows-latest]
@@ -126,7 +126,7 @@ jobs:
         BUILD: ${{ matrix.plan.build }}
       if: contains(matrix.os, 'windows')
 
-    - name: Build on unix
+    - name: Build and test on unix
       run: |
         set -ex
         case "$BUILD" in
@@ -160,7 +160,7 @@ jobs:
         CABALVER: ${{ matrix.plan.cabal-install }}
       if: contains(matrix.os, 'windows') == false
 
-    - name: Build on windows
+    - name: Build and test on windows
       shell: powershell
       run: |
         switch ( "${env:BUILD}" ) {
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -17,14 +17,10 @@ jobs:
 #        os: [ubuntu-latest, macOS-latest, windows-latest]
         os: [ubuntu-latest]
         plan:
-          - { build: stack }
-#          - { build: cabal, ghc: "8.8.3", cabal-install: "latest" }
+          - { resolver: "" }
         include:
           - os: ubuntu-latest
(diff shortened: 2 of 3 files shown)


marcellussiegburg:
> Should it be made much simpler by throwing out all cabal-only stuff? So, switching over to using stack for everything?

I myself am not using cabal directly any more (besides sanity checking before publishing on Hackage; but I am not even calling it directly but using a script for sanity checking and publishing). If we plan on publishing the library on Hackage, it could be helpful to make sanity checks using cabal (at least before publishing), otherwise it is completely redundant in my opinion.

> What's the role of the test argument here [...]?

It is the command what `stack` should be doing in contrast to *only* building (`stack build`) `stack test` also calls the test suite.

> If these commands are for installing dependencies only, there is nothing to test? Or are test suites of the dependencies run as well?

Usually there is nothing to test, but there could be tests for dependencies which, as far as I know may be executed under some circumstances. But I think we do not require such testing. I think this should be changed to `stack build`.

> What is the rationale for the `--only-dependencies` argument here?

This seems to be a copy&paster-error. Interesting that it never got caught, but we turned off testing on other systems than Linux of very early, so maybe there where not many chances.

> And in the above two lines, would it be useful to add --coverage after test?

I am not opinionated in this case. Coverage information could be a helpful indicator. Better would be some specific action that actually rejects the testing if some criteria are not met? Or to have a possibility of integrating some information into a different part (e.g. into the readme file). I do not know if we would have a look at the output of successful builds to check coverage information?

jvoigtlaender:
On this:
> > What's the role of the test argument here [...]?
> 
> It is the command what `stack` should be doing in contrast to _only_ building (`stack build`) `stack test` also calls the test suite.
> 
> > If these commands are for installing dependencies only, there is nothing to test? Or are test suites of the dependencies run as well?
> 
> Usually there is nothing to test, but there could be tests for dependencies which, as far as I know may be executed under some circumstances. But I think we do not require such testing. I think this should be changed to `stack build`.

I had an alternative theory at some point in the last few hours, which was as follows:

Maybe `stack test --only-dependencies` means that dependencies of the **test suite** should be installed? Which could be more than what the library demands. (And installing dependencies of the test suite would install really **everything** needed, since the library itself is also a dependency of the test suite.)

marcellussiegburg:
> Maybe stack test --only-dependencies means that dependencies of the test suite should be installed? Which could be more than what the library demands. (And installing dependencies of the test suite would install really everything needed, since the library itself is also a dependency of the test suite.)

This is in fact the case

PR #194, opened 2023-01-05
don't produce explicit lower bound statement for objects in redundant case

jvoigtlaender:
Issuing the statement `#Obj >= 1` is redundant due to `objectsFact` being included anyway: https://github.com/jvoigtlaender/modelling-tasks-internal/blob/6a20ef26b53af406de46bfe99462e1751a3a2bee/src/Modelling/CdOd/CD2Alloy/Transform.hs#L70-L85

### DIFF
--- a/src/Modelling/CdOd/CD2Alloy/Transform.hs
+++ b/src/Modelling/CdOd/CD2Alloy/Transform.hs
@@ -61,7 +61,7 @@ module umlp2alloy/CD#{index}Module
 
 #{template}
 #{objectsFact}
-#{limitLinks}
+#{sizeConstraints}
 #{loops}
 ///////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Additionally, it is a bit strange that the `#Obj >= ...` statements are put into the https://github.com/jvoigtlaender/modelling-tasks-internal/blob/6a20ef26b53af406de46bfe99462e1751a3a2bee/src/Modelling/CdOd/CD2Alloy/Transform.hs#L90 So probably another change is warranted.

marcellussiegburg:
Agreed, maybe the fact could be renamed to `SizeConstraints`? Or would moving be better (for a more verbose description)?

jvoigtlaender:
Fine either way.

PR #243, opened 2024-02-23
Tasks

marcellussiegburg:
Providing all task configurations used this semester.

Currently the names of the instances corresponds to the task name used this semester within Autotool.

@jvoigtlaender Do you think the proposed PR contains unintentional leaks? Should the structure be changed?

### DIFF
--- a/example/hie.yaml
+++ b/example/hie.yaml
@@ -1,0 +1,9 @@
+cradle:
+  stack:
+    stackYaml: "../stack-example.yaml"
+    components:
+    - path: "../src"
+      component: "modelling-tasks:lib"
+
+    - path: "src"
+      component: "modelling-tasks:lib"
--- a/example/hie.yaml
+++ b/example/hie.yaml
@@ -1,9 +1,9 @@
+---
 cradle:
   stack:
     stackYaml: "../stack-example.yaml"
     components:
-    - path: "../src"
-      component: "modelling-tasks:lib"
-
-    - path: "src"
-      component: "modelling-tasks:lib"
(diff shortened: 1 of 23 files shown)


jvoigtlaender:
I wouldn't be too concerned about leakage. In a few configs, there are tells about the instances (such as which error causes appear in CDs), but even if the same task numbers would be used in another round of the course, actually getting from these configs to a solution strategy (simpler than just solving concrete instances) is probably a rather large step for students.

jvoigtlaender:
Are the capitalization differences in some module names intentional? Such as `EnterAs` here vs. `EnterAS` in the other parts of the repo.

marcellussiegburg:
> Are the capitalization differences in some module names intentional? Such as EnterAs here vs. EnterAS in the other parts of the repo.

Yes.

I tend towards consistent naming for uppercase lettering. Nevertheless I had not made it a priority to rename existing code. Therefore I chose `EnterAs` in line with `CdOd` rather than `EnterAS` (`CdOd` is a good example for choosing the small letter on the second place rather than a capital letter; if both would be capital, it would be hard to guess what `CDOD` is meaning as it looks as if it is a single acronym). Which makes it inconsistent with the (in my opinion) “wrong” version `EnterAS`. But maybe renaming is in order anyway, at least at the Autotool end there is currently no mentioning of activity diagram whatsoever as part of the task name. `EnterAS` should maybe better be `AdEnterAs` to clarify a bit (and take a less prominent name with the namespace of task types). But I think that is out of the scope of this PR.

PR #322, opened 2025-04-03
Replace fmap and <$> with map

### LINKED ISSUE #142: Use `map` instead of `<$>`
As I'm reviewing code, I'm firmly opposed to having `<$>` or `fmap` where `map` would be appropriate. :smile:

(In most places, lists are just a data structure, not the nondeterminism monad or applicative functor, which is what `<$>` is "for".)

nimec01:
This should cover the majority (it's still possible I missed some).

Closes #142


marcellussiegburg:
> (it's still possible I missed some)

Can I ask why that could be? Where not all `<$>` and `fmap` checked? Maybe why? (I would like to estimate how strong the effort was.)

nimec01:
> Can I ask why that could be? Where not all `<$>` and `fmap` checked? Maybe why? (I would like to estimate how strong the effort was.)

I literally searched for all occurrences of `<$>` and `fmap` and checked whether these are called on lists or not. But the Haskell Language Server (or my configuration) has some problems with the project structure, such that it cannot determine the type of value in all files. In that case, I tried my best to determine that myself. That's the "possible I missed some" part.

src/Modelling/ActivityDiagram/Instance.hs
```diff
@@ -122,7 +122,7 @@ parseInstance alloyInstance = do
     <$> getNames scope alloyInstance nodes' "ActionObjectNodes" ComponentName
   let components = enumerateComponents $ toSet nodes'
       names = M.fromList
-        $ zip (nubOrd $ M.elems componentNames) $ pure <$> ['A'..]
+        $ zip (nubOrd $ M.elems componentNames) $ map pure ['A'..]
```
jvoigtlaender: Is this `pure` simply `singleton`?

src/Modelling/ActivityDiagram/Instance.hs
```diff
@@ -230,7 +230,7 @@ getConnections scope alloyInstance ns = do
       (returnX GuardName)
       activityEdges
   let labelMap :: Map GuardName String
-      labelMap = M.fromAscList . zip (S.toAscList triggers) $ pure <$> ['a'..]
+      labelMap = M.fromAscList . zip (S.toAscList triggers) $ map pure ['a'..]
```
jvoigtlaender: Same.

src/Modelling/CdOd/NameCdError.hs
```diff
@@ -776,8 +776,8 @@ nameCdError
   => NameCdErrorConfig
   -> RandT g m (AnyCd, Property, [AnyRelationship String String])
 nameCdError NameCdErrorConfig {..}  = do
-  structuralWeakenings <- shuffleM $ (,)
-    <$> illegalStructuralWeakenings allowedProperties
+  structuralWeakenings <- shuffleM $ map (,)
+    (illegalStructuralWeakenings allowedProperties)
     <*> legalStructuralWeakenings allowedProperties
```
jvoigtlaender: I think here I would prefer seeing a list comprehension explicitly spelling out the Cartesian product.

(The larger point probably being: Just as `<$>` for lists blurs the picture, `<*>` does so even more. If only because there are two possible `Applicative` instances for lists, and just seeing `<*>` requires additional thinking while code-reading, "Is that the zip-list one, or the Cartesian one?" So, I guess I also want all uses of `<*>` on lists specialized away.)

src/Modelling/PetriNet/MatchToMath.hs
```diff
@@ -397,8 +397,8 @@ mathToOutput f pm = paragraph $ do
       f o
       pure ()
   translate $ english ":"
-  itemizeM $ f . fst <$> tokenChangeMath pm
-  itemizeM $ f . snd <$> tokenChangeMath pm
+  itemizeM $ map (f . snd) $ tokenChangeMath pm
+  itemizeM $ map (f . snd) $ tokenChangeMath pm
```
jvoigtlaender, line 401: This seems wrong.

src/Modelling/ActivityDiagram/Instance.hs
marcellussiegburg: Yes, it is!

src/Modelling/PetriNet/Reach/Step.hs
```diff
@@ -100,12 +91,7 @@ levels' n =
   let f _    [] = []
       f done xs =
         let done' = S.union done $ S.fromList $ map fst xs
-            next = M.toList $
-              M.fromList $ do
```
nimec01, line 104: Was there a special reason why `M.toList $ M.fromList` was used here?

jvoigtlaender: What would be the proposed replacement?  That composition does not generally fuse to the identity function.

nimec01: I didn't have in mind that, for example, `M.toList $ M.fromList [(1,'a'),(1,'b')]`  results in `[(1,'b')]` instead of `[(1,'a'),(1,'b')]`.

src/Modelling/PetriNet/Reach/Step.hs
```diff
@@ -100,12 +93,7 @@ levels' n =
...
-                (x, p) <- xs
-                (t, y) <- successors n x
-                guard $ not $ S.member y done'
-                return (y, t : p)
+            next = M.toList $ M.fromList [ (y, t:p) | (x,p) <- xs, (t,y) <- successors n x, not $ S.member y done']
```
marcellussiegburg: Please try to tend to the line length of 80. This line is **very** long.

marcellussiegburg (changes requested):

PR #323, opened 2025-04-03
Rename `avoidAddingSinksForFinals` to `shouldAvoidAddingSinksForFinals`

### LINKED ISSUE #293: Rename `avoidAddingSinksForFinals`
Da es ein `Maybe Bool` ist (statt nur eines `Bool`), ist der Name nicht ganz treffend. Denn etwas nicht zu vermeiden, heißt nicht unbedingt, es zu erzwingen. Also der Unterschied zwischen `Nothing` und `Just False` ist bei der aktuellen Benennung nicht ganz klar.

nimec01:
Closes #293


src/Modelling/ActivityDiagram/SelectPetri.hs
```diff
@@ -136,7 +136,7 @@ data SelectPetriConfig = SelectPetriConfig {
   -- | Option to prevent auxiliary PetriNodes from occurring
   auxiliaryPetriNodeAbsent :: Maybe Bool,
   -- | Avoid having to add new sink transitions for representing finals
-  avoidAddingSinksForFinals :: Maybe Bool,
+  shouldAvoidAddingSinksForFinals :: Maybe Bool,
```
marcellussiegburg: I think this renaming does not yet resolve the name ambiguity. I would propose: `doNotAddSinksForFinals` or swapping the meaning of `Just True` and `Just False` by renaming it to `addSinksForFinals`.

marcellussiegburg (changes requested):

src/Modelling/ActivityDiagram/SelectPetri.hs
jvoigtlaender: How about `sinksPresentForFinals` (and swapped meaning)?

In analogy to existing `Maybe Bool`-options like these: https://github.com/jvoigtlaender/modelling-tasks-internal/blob/13ef35f647a8171b7d98287951ca258cb75648ab/src/Modelling/PetriNet/Types.hs#L743-L745

jvoigtlaender: Could also be `presenceOfSinksForFinals`.

jvoigtlaender: Or indeed `presenceOfSinkTransitionsForFinals` in direct extension of the existing `presenceOfSinkTransitions` above.

PR #328, opened 2025-04-16
Use latest version of check-spelling instead of prerelease

nimec01:
We should probably use the latest (stable) version of `check-spelling` instead of the prerelease. This will resolve the current errors.


.github/workflows/spelling.yml
```diff
@@ -23,12 +23,12 @@ jobs:
     steps:
       - name: check-spelling
         id: spelling
-        uses: check-spelling/check-spelling@prerelease
+        uses: check-spelling/check-spelling@main
```
marcellussiegburg: Actually, using main is also dangerous as changes might come up, as we see with prerelease at the moment). And 0.0.24 (which is the current state of main) is broken as well (see https://github.com/check-spelling/check-spelling/issues/90). Only resolution at the moment would be to pin to the mentioned commit. (in the issue)

marcellussiegburg (changes requested):

patritzenfeld:
This seems to be solved in 0.0.25, so it may be worth revisiting this.

PR #337, opened 2025-07-08
separate names for different Petri instances

marcellussiegburg:
@jvoigtlaender I wanted to send this to discussion before incorporating it. The basic idea is, that different representations of `PetriLike` graphs are separated by name (as their hashes would never overlap useful anyway). But this change requires additional `Data` and `Typeable` constraints in several places (the underlying instances are then used to add parts to the file name). But maybe this is too overcautious?


jvoigtlaender:
The additional `Data` and `Typeable` instances don't bother me.

PR #338, opened 2025-07-09
Move capabilities

patritzenfeld:
Now located at `autotool-capabilities`.

I also had to move [renderSVG](https://github.com/jvoigtlaender/modelling-tasks-internal/blob/dev/src/Modelling/Auxiliary/Diagrams.hs#L609-L616) from Modelling.Auxiliary to prevent a cycling dependency. This function seems to not be used for anything else than the capabilities in this repo, but is it used outside somewhere?


marcellussiegburg:
All `stack*.yaml` files have to be in sync. That is why the checker rejects it. Moreover, both `stack build --stack-yaml=stack-apps.yaml` and `stack build --stack-yaml=stack-examples.yaml` are failing (also locally).

patritzenfeld:
I'm on it

marcellussiegburg:
> I also had to move [renderSVG](https://github.com/jvoigtlaender/modelling-tasks-internal/blob/dev/src/Modelling/Auxiliary/Diagrams.hs#L609-L616) from Modelling.Auxiliary to prevent a cycling dependency. This function seems to not be used for anything else than the capabilities in this repo, but is it used outside somewhere?

Maybe your rip-grep is broken? I see two other places, where it is used (but not in the library parts). However, it is fine to not expose it, if `writeSvg` can be used at the two other places instead. I am not aware of any usages outside this repository.

patritzenfeld:
> > I also had to move [renderSVG](https://github.com/jvoigtlaender/modelling-tasks-internal/blob/dev/src/Modelling/Auxiliary/Diagrams.hs#L609-L616) from Modelling.Auxiliary to prevent a cycling dependency. This function seems to not be used for anything else than the capabilities in this repo, but is it used outside somewhere?
> 
> Maybe your rip-grep is broken? I see two other places, where it is used (but not in the library parts). However, it is fine to not expose it, if `writeSvg` can be used at the two other places instead. I am not aware of any usages outside this repository.

Isn't that another function that just happens to have the same name? (from `Diagrams.Backend.SVG`)

marcellussiegburg (approved):

PR #339, opened 2025-07-09
derive Data for Net

horriblename:
I had to derive Data for `GraphvizCommand`, which I'm not sure is a good idea

### DIFF
--- a/src/Modelling/PetriNet/Reach/Reach.hs
+++ b/src/Modelling/PetriNet/Reach/Reach.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TupleSections #-}
 
 {-|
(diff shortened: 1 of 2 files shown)


src/Modelling/PetriNet/Reach/Reach.hs
```diff
@@ -20,6 +21,7 @@ import qualified Data.Set                         as S (toList)
 import Capabilities.Cache               (MonadCache)
 import Capabilities.Diagrams            (MonadDiagrams)
 import Capabilities.Graphviz            (MonadGraphviz)
+import Data.Data                        (Data(..))
```
marcellussiegburg: I do not see, why `Data` is not enough here. Why do you need constructors?

src/Modelling/PetriNet/Reach/Type.hs
```diff
@@ -22,6 +22,7 @@ import qualified Data.Set                         as S (fromList, map)
 import Modelling.Auxiliary.Common       (parseInt, skipSpaces)
 
 import Control.Monad                    (void)
+import Data.Data (Data (..))
```
marcellussiegburg: Same here

marcellussiegburg (changes requested):

marcellussiegburg:
> I had to derive Data for `GraphvizCommand`, which I'm not sure is a good idea

It surely should be avoided to define custom instances of data types of libraries. But if `Data` instances are required for data types consisting of `GraphvizCommand` there is no way around it.

src/Modelling/PetriNet/Reach/Reach.hs
horriblename: you're right, they're not needed

horriblename: removed

PR #343, opened 2025-07-21
refactor out components to generate a petri simulation

horriblename:
as discussed in https://github.com/fmidue/ba-pei-ching/pull/3#discussion_r2203461996

Refactors out components that are commonly used to generate a "simulation" i.e. an initial net + transitions + the final state after transitions

### DIFF
--- a/src/Modelling/PetriNet/Reach/Reach.hs
+++ b/src/Modelling/PetriNet/Reach/Reach.hs
@@ -291,7 +291,7 @@ isNoLonger maybeMaxLength ts =
         ]
 
 data ReachInstance s t = ReachInstance {
-  netGoal           :: NetGoalInstance s t,
+  netGoal           :: NetGoal s t,
   minLength         :: Int,
   noLongerThan      :: Maybe Int,
   showGoalNet       :: Bool,
@@ -300,7 +300,7 @@ data ReachInstance s t = ReachInstance {
   withMinLengthHint :: Maybe Int
   } deriving (Generic, Read, Show, Typeable, Data)
 
-data NetGoalInstance s t = NetGoalInstance {
+data NetGoal s t = NetGoal {
   drawUsing         :: GraphvizCommand,
   petriNet          :: Net s t,
   goal              :: State s
@@ -313,7 +313,7 @@ bimapReachInstance
   -> ReachInstance s t
   -> ReachInstance a b
 bimapReachInstance f g ReachInstance {..} = ReachInstance {
-    netGoal           = bimapNetGoalInstance f g netGoal,
+    netGoal           = bimapNetGoal f g netGoal,
     minLength         = minLength,
     noLongerThan      = noLongerThan,
     showGoalNet       = showGoalNet,
@@ -322,13 +322,13 @@ bimapReachInstance f g ReachInstance {..} = ReachInstance {
     withMinLengthHint = withMinLengthHint
     }
 
-bimapNetGoalInstance
+bimapNetGoal
   :: (Ord a, Ord b)
   => (s -> a)
   -> (t -> b)
-  -> NetGoalInstance s t
-  -> NetGoalInstance a b
-bimapNetGoalInstance f g NetGoalInstance {..} = NetGoalInstance {
+  -> NetGoal s t
+  -> NetGoal a b
+bimapNetGoal f g NetGoal {..} = NetGoal {
     drawUsing = drawUsing,
     goal      = mapState f goal,
     petriNet  = bimapNet f g petriNet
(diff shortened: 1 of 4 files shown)


jvoigtlaender:
I don't think `Simulation` is the best name.

And the ` GraphvizCommand` stuff shouldn't be part of these new types.

horriblename:
the other name I had in mind was Execution, I don't know what else I can name it...

jvoigtlaender:
@marcellussiegburg, do you have a suggestion? Is it not mainly a `NetConfig` or so?

horriblename:
> And the  GraphvizCommand stuff shouldn't be part of these new types.

as I understand, the GraphvizCommand is needed to verify that the generated graph is drawable: https://github.com/jvoigtlaender/modelling-tasks-internal/blob/7ed32cfd67d1bd2d37513959db35f93095c0af61/src/Modelling/PetriNet/Reach/Reach.hs#L404

notably, `isPetriDrawable` requires a GraphvizCommand, and might fail if graphviz fails to layout the net

jvoigtlaender:
Okay, so in that sense it is "part of configuring the net"?

Well, makes some sense.

Still, it does not mean that it absolutely needs to be part of that "factored out" record. After all, generation functions could also take the `GraphvizCommand`(s) to check for drawability as separate arguments.

I will defer to @marcellussiegburg's judgement here.

marcellussiegburg:
It is just one component more than just the net, i.e. the target. How about `ReachGoalConfig` or `NetGoalConfig`?

About the `GraphvizCommand`: It does seem that having it as part of the data type seems unintuitive regarding the intention. But it might be convenient to have it there anyway. The question is: What is the probable usage of this data type elsewhere? Would it be feasible to have some situation where no net would be drawn at all? If not, it could be part of this data type. But still it might be possible in other cases to configure even more drawing settings (in the future)? Thus, it would make more sense to reuse `Modelling.PetriNet.Types`' `Drawable`. (At the cost of having to define all those parameters and introducing checkers for all config options to prevent invalid options for current tasks.)

src/Modelling/PetriNet/Reach/Reach.hs
```diff
@@ -291,96 +291,124 @@ isNoLonger maybeMaxLength ts =
...
   minTransitionLength :: Int,
   postconditionsRange :: (Int, Maybe Int),
-  preconditionsRange  :: (Int, Maybe Int),
-  printSolution       :: Bool,
-  rejectLongerThan    :: Maybe Int,
-  showLengthHint      :: Bool,
-  showMinLengthHint   :: Bool,
-  showTargetNet       :: Bool
+  preconditionsRange  :: (Int, Maybe Int)
   }
```
marcellussiegburg: The remaining configuration parameters describe a Petri net rather than a Simulation.

src/Modelling/PetriNet/Reach/Reach.hs
```diff
@@ -291,96 +291,124 @@ isNoLonger maybeMaxLength ts =
...
+data NetGoalConfig s t = NetGoalConfig {
+  simulation        :: NetGoalInstance s t,
   minLength         :: Int,
   noLongerThan      :: Maybe Int,
-  petriNet          :: Net s t,
   showGoalNet       :: Bool,
   showSolution      :: Bool,
   withLengthHint    :: Maybe Int,
   withMinLengthHint :: Maybe Int
   } deriving (Generic, Read, Show, Typeable, Data)
```
marcellussiegburg: What happened here? Why did the instance become a config? Maybe I am not in the right picture what the reuse tries to do? Should it be used as an instance elsewhere? Why was it necessary to rename it in the first place? I think it is wise to discuss the changes first? I cannot see the intention of this change.

marcellussiegburg (changes requested):

src/Modelling/PetriNet/Reach/Reach.hs
horriblename: I believe I typed in the wrong name when trying to rename SimulationConfig to NetGoalConfig. now fixed

src/Modelling/PetriNet/Reach/Reach.hs
```diff
@@ -403,23 +431,37 @@ generateReach conf seed = do
...
-    noLongerThan      = rejectLongerThan conf,
-    petriNet          = petri,
-    showGoalNet       = showTargetNet conf,
-    showSolution      = printSolution conf,
-    withLengthHint    =
-      if showLengthHint conf then Just $ maxTransitionLength conf else Nothing,
-    withMinLengthHint =
-      if showMinLengthHint conf then Just $ minTransitionLength conf else Nothing
+
+  return $ NetGoalInstance {
```
jvoigtlaender: What’s the motivation to switch from pure to return here?

jvoigtlaender:
Is netGoalConf :: NetGoalConfig somehow following a convention from elsewhere? Why is it not netGoalConfig :: NetGoalConfig?

jvoigtlaender:
It seems …Instance is usually reserved for actual task instances. So, NetGoalInstance should probably just be NetGoal.

horriblename:
> Is netGoalConf :: NetGoalConfig somehow following a convention from elsewhere? Why is it not netGoalConfig :: NetGoalConfig?

I saw this https://github.com/jvoigtlaender/modelling-tasks-internal/blob/7ed32cfd67d1bd2d37513959db35f93095c0af61/src/Modelling/ActivityDiagram/MatchAd.hs#L83 (there are a few other examples) so I named it that way, but there also quite a few that's named ...Config. I'll change it to netGoalConfig

src/Modelling/PetriNet/Reach/Reach.hs
```diff
@@ -403,23 +431,37 @@ generateReach conf seed = do
...
+    netGoal           = netGoal,
+    minLength         = minTransitionLength (netGoalConfig conf),
+    noLongerThan      = rejectLongerThan conf,
+    showGoalNet       = showTargetNet conf,
+    showSolution      = printSolution conf,
+    withLengthHint    =
+      if showLengthHint conf then Just $ maxTransitionLength (netGoalConfig conf) else Nothing,
+    withMinLengthHint =
+      if showMinLengthHint conf then Just $ minTransitionLength (netGoalConfig conf) else Nothing
+    }
```
marcellussiegburg: This might benefit a lot from the use of `RecordWildCards`.

PR #345, opened 2025-07-30
derive Data NameCdErrorInstance

horriblename:
depends on https://github.com/fmidue/output-blocks/pull/31

I will update stack.yaml after that gets merged


src/Modelling/CdOd/NameCdError.hs
```diff
@@ -213,7 +215,7 @@ $(deriveJSON defaultOptions {fieldLabelModifier = upperToDash} ''NameCdErrorAnsw
 data Reason
   = Custom (Map Language String)
   | PreDefined Property
-  deriving (Eq, Generic, Ord, Read, Show)
+  deriving (Eq, Generic, Ord, Read, Show, Data)
```
marcellussiegburg: @horriblename Would you mind changing this into alphabetical order (below as well)?

PR #346, opened 2025-08-01
add deriving Read to SelectASInstance

guard-who:
I need SelectASInstance deriving `Read`

### DIFF
--- a/src/Modelling/ActivityDiagram/SelectAS.hs
+++ b/src/Modelling/ActivityDiagram/SelectAS.hs
@@ -95,7 +95,7 @@ data SelectASInstance = SelectASInstance {
   drawSettings :: PlantUmlConfig,
   showSolution :: Bool,
   addText :: Maybe (Map Language String)
-} deriving (Generic, Show, Eq)
+} deriving (Eq, Generic, Read, Show)
 
 data SelectASConfig = SelectASConfig {
   adConfig :: AdConfig,


marcellussiegburg:
I am not sure, what you are trying to achieve, but this involves just requiring Read for this single Instance? I am asking because there is at least also `src/Modelling/ActivityDiagram/SelectPetri.hs` where the instance does not provide a Read instance. But others won't be required?

guard-who:
I am trying to inject real-world concepts, so I need to call here [https://github.com/jvoigtlaender/modelling-tasks-internal/blob/43c19537fe5be62bbc3a743114b4306d9431f0a7/app/selectASTaskDemo.hs#L24](https://github.com/jvoigtlaender/modelling-tasks-internal/blob/43c19537fe5be62bbc3a743114b4306d9431f0a7/app/selectASTaskDemo.hs#L24) to generate an instance and write it into a file. Then I move the instance file to the repo [llm-task-wrapper](https://github.com/fmidue/llm-task-wrappers). I need to read the instance and process it at here [https://github.com/fmidue/llm-task-wrappers/blob/9c15304c83aa31b7be122d8acd53e15af17dff45/app/selectAS.hs#L22](https://github.com/fmidue/llm-task-wrappers/blob/9c15304c83aa31b7be122d8acd53e15af17dff45/app/selectAS.hs#L22). So I only need `SelectASInstance` deriving `Read`.

But I may need to add some `write` operations, e.g. [https://github.com/jvoigtlaender/modelling-tasks-internal/blob/43c19537fe5be62bbc3a743114b4306d9431f0a7/app/selectASTaskDemo.hs#L24C7-L25C17](https://github.com/jvoigtlaender/modelling-tasks-internal/blob/43c19537fe5be62bbc3a743114b4306d9431f0a7/app/selectASTaskDemo.hs#L24C7-L25C17). It generates and print the task, but it doesn't save the task instance in a file. Currently, I am not sure if there are better choices. But at least, `SelectASInstance` deriving `Read` is needed and I don't need others. Do you think adding deriving `READ` for  `src/Modelling/ActivityDiagram/SelectPetri.hs` to keep consistency is better?

marcellussiegburg:
I have no access to the other Repository, so I cannot see what is happening there. But if amendment to `selectASTaskDemo` is required as well. It could be part of this pull request? But maybe you would not require the output of the task and interaction with it, so you would likely need to create your own wrapper, that calls `selectAS defaultSelectASConfig` and saves the result to file? Let me know. If you just need `defaultSelectASConfig` that is fine. I just wanted to prevent having multiple pull requests just handling a single deriving of an instance.

guard-who:
After double check, As Injecting concepts into `SelectPetri` is not so interesting, it is not considered yet. All other interesting Instances have derived from `Read` , so I think I only need to derive `Read` for `SelectASInstance` in current.

As for writing instances into files, I can indeed build executables in my wrapper, which can reduce non-determined modifications here.

PR #351, opened 2025-08-27
derive Data MatchInstance

horriblename:
needed for https://github.com/fmidue/ba-pei-ching/pull/11

### DIFF
--- a/src/Modelling/PetriNet/MatchToMath.hs
+++ b/src/Modelling/PetriNet/MatchToMath.hs
@@ -7,6 +7,8 @@
 {-# Language QuasiQuotes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TupleSections #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE StandaloneDeriving #-}
 
 module Modelling.PetriNet.MatchToMath (
   GraphToMathInstance,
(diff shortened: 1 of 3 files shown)


src/Modelling/PetriNet/MatchToMath.hs
```diff
@@ -192,6 +194,8 @@ data MatchInstance a b = MatchInstance {
   }
   deriving (Functor, Generic, Read, Show)
 
+deriving instance (Data a, Data b) => Data (MatchInstance a b)
+
```
marcellussiegburg, line 195: Why is the Standalone deriving required here? This could just be part of the deriving in the line above?

src/Modelling/PetriNet/Types.hs
```diff
@@ -889,7 +890,9 @@ data DrawSettings = DrawSettings {
   withTransitionNames  :: Bool,
   with1Weights         :: Bool,
   withGraphvizCommand  :: GraphvizCommand
-  } deriving (Generic, Read, Show)
+  } deriving (Data, Generic, Read, Show)
+
+deriving instance Data GraphvizCommand
```
marcellussiegburg, line 895: This is Dangerous. There is already an instance in `src/Modelling/PetriNet/Reach/Reach.hs`. There should only live one such instance in the repo.

marcellussiegburg (changes requested):

src/Modelling/PetriNet/Types.hs
horriblename: I've moved the instance from `src/Modelling/PetriNet/Reach/Reach.hs` to this file, as that one imports from `Modelling.PetriNet.Types` transitively.

PR #407, opened 2025-09-26
add flag to control input hint in task text

horriblename:


src/Modelling/CdOd/DifferentNames.hs
```diff
@@ -350,6 +355,8 @@ toTaskText
   -> LangM m
 toTaskText path task = do
   specialToOutputCapable (toTaskSpecificText path task) (taskText task)
+  when (inputHelp task) $
```
jvoigtlaender: Why do you need that `Bool` data extracted from the instance? Why can't it simply be a parameter that is passed here from `differentNamesTask`?

src/Modelling/CdOd/DifferentNames.hs
```diff
@@ -327,11 +329,12 @@ data DifferentNamesTaskTextElement
...
   :: (MonadCache m, MonadDiagrams m, MonadGraphviz m, MonadThrow m, OutputCapable m)
-  => FilePath
+  => Bool
+  -> FilePath
   -> DifferentNamesInstance
   -> LangM m
-differentNamesTask path task = do
-  toTaskText path task
+differentNamesTask inputHelp path task = do
+  toTaskText inputHelp path task
```
marcellussiegburg: I see the phrasing of the variable Name as problematic for two reasons:
1. the variable that is influenced is called `inputHintText`, why is this called `inputHelp`?
2. the variable name of a Boolean should imply its containing Boolean values (or in other words: The line starting with `when` (further below) should sound naturally when being read)

I propose: `showInputHint`

marcellussiegburg (changes requested):

src/Modelling/CdOd/SelectValidCd.hs
```diff
@@ -315,14 +321,18 @@ defaultSelectValidCdTaskText = [
     german [i|Betrachten Sie die folgenden Klassendiagrammkandidaten:|],
   Special CdCandidates,
   Paragraph $ singleton $ Translated $ translations $ do
-    english [i|Which of these class diagram candidates are valid class diagrams?
-Please state your answer by giving a list of numbers, indicating all valid class diagrams.|]
-    german [i|Welche dieser Klassendiagrammkandidaten sind gültige Klassendiagramme?
-Bitte geben Sie Ihre Antwort in Form einer Liste von Zahlen an, die alle gültigen Klassendiagramme enthält.|],
+    english [i|Please state your answer by giving a list of numbers, indicating all valid class diagrams.|]
+    german [i|Welche dieser Klassendiagrammkandidaten sind gültige Klassendiagramme?|]
```
marcellussiegburg: These translations do not match. This is clearly wrong. The text that was omitted needs to be used here.

marcellussiegburg (changes requested):

marcellussiegburg (approved):

PR #414, opened 2025-10-02
Use custom data type and `collapsible` for extra text

### LINKED ISSUE #341: Use `collapsed` for `extraText`s
From https://github.com/fmidue/output-blocks/blob/e97c5f06a17c5b99f0bed2a733bfbf974304adf3/output-blocks/src/Control/OutputCapable/Blocks.hs#L511-L516 we currently have this function:
```haskell
collapsed
  :: OutputCapable m
  => Bool
  -> State (Map Language String) ()
  -> LangM m
  -> LangM m
```
(but see https://github.com/fmidue/output-blocks/issues/34)

nimec01:
Closes #341

Some changes might be necessary after https://github.com/fmidue/output-blocks/issues/34.

### DIFF
--- a/src/Modelling/Auxiliary/Output.hs
+++ b/src/Modelling/Auxiliary/Output.hs
@@ -1,7 +1,9 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE DeriveDataTypeable #-}
 -- | This module provides common skeletons for printing tasks
 module Modelling.Auxiliary.Output (
+  ExtraText(..),
   addPretext,
   checkTaskText,
   directionsAdvice,
@@ -11,18 +13,19 @@ module Modelling.Auxiliary.Output (
   uniform,
   ) where
 
-import qualified Data.Map                         as M (empty, insert)
+import qualified Data.Map                         as M (empty, insert, fromList)
 
 import Control.Monad.State (put)
 import Control.OutputCapable.Blocks     (
   GenericOutputCapable (paragraph),
-  Language,
+  Language(..),
   LangM,
   LangM',
   OutputCapable,
   english,
   german,
   translate,
+  collapsed,
   )
 import Control.OutputCapable.Blocks.Type (
   SpecialOutput,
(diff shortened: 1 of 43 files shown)


src/Modelling/Auxiliary/Output.hs
```diff
@@ -113,6 +117,17 @@ checkTaskText taskText
...
+  | Collapsible Bool (Map Language String)
+  deriving (Data, Eq, Read, Show)
+
+extra :: OutputCapable m => ExtraText -> LangM m
+extra NoExtraText = pure ()
+extra (Static textMap) = paragraph $ translate $ put textMap
+extra (Collapsible defaultState textMap) =
+  collapsed
+    defaultState
+    (put $ M.fromList [(English, "Additional information"), (German, "Zusätzliche Informationen")])
```
jvoigtlaender: Maybe the title should also be parametrized, by adding  another argument to the `Collapsible` constructor.

jvoigtlaender:
@nimec01, is this ready for review/inclusion?

nimec01:
> @nimec01, is this ready for review/inclusion?

Yes.

src/Modelling/Auxiliary/Output.hs
```diff
@@ -113,6 +117,17 @@ checkTaskText taskText
     usedElements = concatMap (concatMap singleton) taskText
     allElements = [minBound ..]
 
-extra :: OutputCapable m => Maybe (Map Language String) -> LangM m
-extra (Just extraMap) = paragraph $ translate $ put extraMap
-extra _ = pure ()
+data ExtraText
+  = NoExtraText
+  | Static (Map Language String)
+  | Collapsible Bool (Map Language String) (Map Language String)
```
marcellussiegburg: This should get Haddock comments for the parameters! I know it is just reusing functions of `output-blocks` below. But linking to the documentation is what should be done at least in my opinion, or even better: explain it here. It becomes clear when reading below. But a user should be able to find the information when looking into haddock documentation (https://fmidue.github.io/modelling-tasks/).

marcellussiegburg (changes requested):

marcellussiegburg (approved):

PR #460, opened 2025-12-01
Refine filtering out isolated nodes on reach generation

jvoigtlaender:
- also, remove dead code in the generation module

### DIFF
--- a/src/Modelling/PetriNet/Reach/Deadlock.hs
+++ b/src/Modelling/PetriNet/Reach/Deadlock.hs
@@ -52,7 +52,6 @@ import Modelling.PetriNet.Reach.Type (
   TransitionsList (TransitionsList),
   bimapNet,
   example,
-  hasIsolatedNodes,
   )
 
 import Control.Applicative              (Alternative)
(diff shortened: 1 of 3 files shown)


jvoigtlaender:
This is doing what https://github.com/fmidue/modelling-tasks/commit/6ba60297804f543c43d243314621a0f89838fc9d in #379 purported to do.

What I'm not certain about is the changes to the type signatures in `Deadlock.hs` and `Reach.hs`. I simply followed the compiler on those. In particular, I didn't even try to understand why sometimes it wants `Alternative m` and sometimes `MonadPlus m`. And I don't know whether the constraints added to these functions will be fine at their use sites (in terms of `autotool-capabilities`, for example).

jvoigtlaender:
@marcellussiegburg, can you check this, and in particular also Codex's complaint above?

jvoigtlaender:
There are two test failures: https://github.com/fmidue/modelling-tasks/actions/runs/19824798803/job/56795353954?pr=460#step:10:2179

I suppose they would also have occurred with https://github.com/fmidue/modelling-tasks/commit/6ba60297804f543c43d243314621a0f89838fc9d if that hadn't applied the "refactoring" to dead code.

PR #477, opened 2025-12-15
fix unit tests for dirCommand usage instead of undirCommand

### LINKED ISSUE #476: Display of object diagrams
With longer names for classes, and in particular for relationships, we run into cases where at some point even hovering does not really help anymore:

Image

Worse:

Image

Tweaks need to be found to make this more consumable.

guard-who:
variant 1:
image

variant 2:
image

`dirCommand` has a better layout than `undirCommand`, but not so compact as `undirCommand`. Using `dirCommand` can avoid extra text to describe mappings which possibly distract students.


jvoigtlaender:
Adding `dirCommand` to the import list here is superfluous. What you need to do, though, is to fix the unit tests: https://github.com/fmidue/modelling-tasks/actions/runs/20235293139/job/58088272201

guard-who:
Unit tests have been fixed.

PR #490, opened 2025-12-17
More clear records for configurations of task 14-16

guard-who:


jvoigtlaender:
Please create more sanitized pull requests, without so much unnecessary "back story". Probably by not re-using old/stale branches so much.

About the changes of this PR here itself:
- It's better to indeed put `anonymousObjectProportion = 0 % 1` into `task2025_14` and then do explicit record updates to set `anonymousObjectProportion = 1 % 1` in the lines for `task2025_15` and `task2025_16`.
- Record in comments also whether or not these three tasks share the same underlying instances.
- Record in comments also whether or not the last two tasks share the same concept injections (instead of two different injection runs).
- Record how many underlying instances were generated, what the time usage for that was on the cluster, and for the injection runs also what the "token costs" were.

guard-who:
> and for the injection runs also what the "token costs" were.

Could you help me check the total approximate "token costs" last Thursday and Friday for the OpenAI API key you shared with me? I can't access it directly via the API key.

jvoigtlaender:
The Dashboard shows me this:

image

image

----

image

image

----

image

image

PR #493, opened 2025-12-17
DifferentNames: Do not discard mappings with unused class diagram links

### LINKED ISSUE #492: Incorrect feedback output with `multipleChoice`
At least in `differentNamesEvaluate`, the final grading and feedback does not punish/handle mappings, which assign an unused relationship from the class diagram, correctly. Instead, they are thrown away, leading to no punishment for such mistakes and feedback claiming that all given answers are correct.

This could also be an issue elsewhere, so we should make sure all tasks contained here handle similar mistakes correctly.

patritzenfeld:
This caused feedback to assert

**All indicated mappings are correct?**
```
Yes.
```

even when wrong mappings were present.

### DIFF
--- a/src/Modelling/CdOd/DifferentNames.hs
+++ b/src/Modelling/CdOd/DifferentNames.hs
@@ -570,9 +570,9 @@ differentNamesEvaluation task cs = do
       mStripped = BM.mapMonotonicR stripName $ nameMapping $ mapping task
       -- Swap answer tuples around if necessary
       -- The preceding syntax check guarantees only valid pairs can be submitted here
-      readMapping pair@(left,_)
-        | BM.member left mStripped = pair
-        | otherwise = swap pair
+      readMapping pair@(_, right)
+        | BM.member right mStripped = swap pair
+        | otherwise = pair
       what = translations $ do
         german "Zuordnungen"
         english "mappings"
--- a/src/Modelling/CdOd/DifferentNames.hs
+++ b/src/Modelling/CdOd/DifferentNames.hs
@@ -227,8 +227,8 @@ checkDifferentNamesInstance DifferentNamesInstance {..}
       but currently "#{x}" is among both.
       |]
   | let strippedODMapping = BM.map
-      stripNumericPeriod
-      $ fromNameMapping mapping,
+          stripNumericPeriod
+          $ fromNameMapping mapping,
     any (`BM.member` strippedODMapping) strippedLinks
   = Just [iii|
       Pairs given in mapping must follow this order:
--- a/src/Modelling/CdOd/DifferentNames.hs
+++ b/src/Modelling/CdOd/DifferentNames.hs
@@ -35,6 +35,7 @@ import qualified Data.Bimap                       as BM (
   fromList,
   keys,
   lookup,
+  mapMonotonic,
   mapMonotonicR,
   member,
   toAscList,
@@ -220,10 +221,18 @@ checkDifferentNamesInstance DifferentNamesInstance {..}
       i.e., for which an association in the Class diagram exists
       but not a link in the Object diagram.
       |]
-  | (x:_) <- nubOrd links `intersect` nubOrd associations
+  | (x:_) <- nubOrd strippedLinks `intersect` nubOrd associations
   = Just [iii|
       Link names and association names must be disjoint
       but currently "#{x}" is among both.
+      |]
+  | strippedODMapping <- BM.mapMonotonic
+      stripNumericPeriod
+      $ fromNameMapping mapping,
+    any (`BM.member` strippedODMapping) strippedLinks
+  = Just [iii|
+      Pairs given in mapping must follow this order:
+      (CD association, OD link)
       |]
   | otherwise
   = checkObjectDiagram oDiagram
@@ -231,6 +240,7 @@ checkDifferentNamesInstance DifferentNamesInstance {..}
   where
     associations = associationNames cDiagram
     links = linkLabels oDiagram
+    strippedLinks = map stripNumericPeriod links
 
 data DifferentNamesConfig
   = DifferentNamesConfig {
--- a/src/Modelling/CdOd/DifferentNames.hs
+++ b/src/Modelling/CdOd/DifferentNames.hs
@@ -37,7 +37,6 @@ import qualified Data.Bimap                       as BM (
   lookup,
   mapMonotonicR,
   member,
-  memberR,
   toAscList,
   )
 import qualified Data.Map                         as M (
@@ -561,9 +560,8 @@ differentNamesEvaluation task cs = do
       mStripped = BM.mapMonotonicR stripName $ nameMapping $ mapping task
       -- Swap answer tuples around if necessary
       -- The preceding syntax check guarantees only valid pairs can be submitted here
-      readMapping pair@(left,right)
-        | BM.member left mStripped ||
-          BM.memberR right mStripped = pair
+      readMapping pair@(left,_)
+        | BM.member left mStripped = pair
         | otherwise = swap pair
       what = translations $ do
         german "Zuordnungen"
--- a/src/Modelling/CdOd/DifferentNames.hs
+++ b/src/Modelling/CdOd/DifferentNames.hs
@@ -36,7 +36,8 @@ import qualified Data.Bimap                       as BM (
   keys,
   lookup,
   mapMonotonicR,
-  pairMember,
+  member,
+  memberR,
   toAscList,
   )
 import qualified Data.Map                         as M (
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Is that really the correct check? Won't it also let things like `(x,x)` through?

jvoigtlaender:
Why isn't it left at the conjunctive checks, but the lookup being with respect to the actual class and object diagrams (the relationship/link names appearing in them) instead of with respect to the sample solution?

patritzenfeld:
> Is that really the correct check? Won't it also let things like `(x,x)` through?

The syntax check would already reject something like that, albeit with a "weird" remark:

`The mapping '(x,x)' uses a non-existing identifier.`

patritzenfeld:
> Why isn't it left at the conjunctive checks, but the lookup being with respect to the actual class and object diagrams (the relationship/link names appearing in them) instead of with respect to the sample solution?

I can do that instead, it would just take a bit more digging through the instance. I just went with the easiest and smallest change I could find.

jvoigtlaender:
Okay, different question then (I haven't looked at the syntax check): Given that the syntax check has run before, when would the new version of readMapping ever return Nothing?

patritzenfeld:
It seems like that can't happen. So, the function essentially only exists to swap the tuple components around if necessary.
The syntax check already makes sure that:

1. no unknown identifiers can be used (includes doubled identifiers like `(x,x)`)
2. no duplicate mappings are included (i.e. `[(x,1), (x,2)]`)

By doing what you proposed earlier with the complete diagram names.

I have to admit this is confusing to me, because making this function return a Maybe value seems to hint the intent was to explicitly filter those pairs we want to allow now, because that is the only thing that could have arrived at the Nothing guard beforehand.

jvoigtlaender:
Well, your 1. and 2. do not mention the case where a pair `(x,y)` is input, with `x` and `y` different but both coming from the same diagram (instead of one coming from the class diagram and the other coming from the object diagram).

Will such pairs survive the syntax check?

patritzenfeld:
No, I wasn't clear enough about 1. It checks if one of the given pair is in the CD and one is in the OD. So (x,y) or (1,2) would also not pass.

jvoigtlaender:
No idea whether #508 is related.

patritzenfeld:
How am I to proceed with this? I can do as you initially suggested (inspect the actual diagrams in the semantics check), but that would mean redundancy in the syntax and semantics checks, as the syntax check already does the same.

If we go the no redundancy route, then the discussed function would simply be replaced by a tuple component swap without Maybe being involved.

jvoigtlaender:
The latter seems to make the most sense (relying on the syntax check having run before, or the form inputs preventing syntax errors in the first place).

src/Modelling/CdOd/DifferentNames.hs
```diff
@@ -569,6 +558,12 @@ differentNamesEvaluation task cs = do
   let csStripped = map (bimap stripName stripName) cs
       -- Strip periods from the mapping's link labels (second element of each pair)
       mStripped = BM.mapMonotonicR stripName $ nameMapping $ mapping task
+      -- Swap answer tuples around if necessary
+      -- The preceding syntax check guarantees only valid pairs can be submitted here
+      readMapping pair =
+        if BM.pairMember pair mStripped
+        then pair
+        else swap pair
```
jvoigtlaender: Is it intentional that this decides about swapping by checking whether the submitted pair occurs in the correct solution?

That is, let's say the correct solution is given as `[(1,x), (2,y)]` (numbers left, letters right). If a student submits `(2,x)`, it will be swapped to `(x,2)`, so into the "wrong order".

patritzenfeld: That's not intentional, I forgot about that case. I will change this to look at the first component of each entry instead.

patritzenfeld: Ended up checking both components (for cases where some CD names are unused).

src/Modelling/CdOd/DifferentNames.hs
```diff
@@ -569,6 +559,12 @@ differentNamesEvaluation task cs = do
   let csStripped = map (bimap stripName stripName) cs
       -- Strip periods from the mapping's link labels (second element of each pair)
       mStripped = BM.mapMonotonicR stripName $ nameMapping $ mapping task
+      -- Swap answer tuples around if necessary
+      -- The preceding syntax check guarantees only valid pairs can be submitted here
+      readMapping pair@(left,right)
+        | BM.member left mStripped ||
+          BM.memberR right mStripped = pair
```
jvoigtlaender: What's a situation where the `|| BM.memberR right mStripped` becomes relevant, i.e., where a `(left,right)` has survived the syntax check but `BM.member left mStripped /= BM.memberR right mStripped`?

patritzenfeld: When `usesEveryRelationshipName` is not set to `Just True`. `mStripped` is derived from the solution and that solution mapping can be given in arbitrary order, e.g. `[(1,x),(2,y)]` or `[(x,1),(y,2)]` (of course the order is always the same when generated, but a direct task exists in Autotool).

So it could be the case, that the left component of an answer tuple is not in the mapping at all, if it went unused in the OD, but still got through the syntax check, because it does assign two valid links to each other.

jvoigtlaender: Oh, that the solution mapping in the instance can be given in any pair order seems to be wrong. There is even a code comment just a few lines above explicitly stating that the second pair components of the given solution mapping should be the link labels. (Or else the `stripName` would not touch all the strings it needs to touch!)

So, that first means there should be an instance check to prevent what you describe (someone swapping tuples in the solution mapping).

Second, what I do think is the case that in the submission by the student the pair component order could be arbitrary, even mixed. But for that situation I still do not see how one needs to check both for `BM.member left mStripped` and for `BM.memberR right mStripped` in order to decide whether to swap or not.

jvoigtlaender: > because it does assign two valid links to each other

I thought the syntax checker prevented something like `(1,2)`, i.e., assigning two links to each other instead of a link and a relationship or a relationship and a link.

patritzenfeld: My bad, I meant something like (1,z) (where z is not in the mapping), not literally two links from the same diagram.

patritzenfeld: Maybe another thing that's unclear: In the syntax check, it is actually confirmed that all pair components are part of either diagram by directly checking the structures in the instance. This later check only looks at the solution, so if a relationship is not used, then it is not recorded in that solution.

jvoigtlaender: > I meant something like (1,z) (where z is not in the mapping)

But in that case, what does it matter whether `(1,z)` or `(z,1)` is passed on to the `multipleChoice` function for grading? It's a tuple going to be judged as wrong (and accordingly punished) either way.

patritzenfeld: You are correct. I thought this would still trigger the "All indicated... are correct?" message, but that is not the case after removing the filtering in modelling-tasks. So I guess just checking the first component is already enough.

patritzenfeld: > Oh, that the solution mapping in the instance can be given in any pair order seems to be wrong. There is even a code comment just a few lines above explicitly stating that the second pair components of the given solution mapping should be the link labels. (Or else the stripName would not touch all the strings it needs to touch!)

I will also add this. I just confirmed this is not enforced by config checkers.

src/Modelling/CdOd/DifferentNames.hs
```diff
@@ -570,9 +570,9 @@ differentNamesEvaluation task cs = do
       mStripped = BM.mapMonotonicR stripName $ nameMapping $ mapping task
       -- Swap answer tuples around if necessary
       -- The preceding syntax check guarantees only valid pairs can be submitted here
-      readMapping pair@(left,_)
-        | BM.member left mStripped = pair
-        | otherwise = swap pair
+      readMapping pair@(_, right)
+        | BM.member right mStripped = swap pair
```
jvoigtlaender: One could argue there is no point of doing this lookup in `mstripped`. Since it is only checking the first pair component, which is unaffected by `BM.mapMonotonicR` above, the check could be done on the original mapping.

src/Modelling/CdOd/DifferentNames.hs
```diff
@@ -222,17 +221,26 @@ checkDifferentNamesInstance DifferentNamesInstance {..}
...
       but not a link in the Object diagram.
       |]
-  | (x:_) <- nubOrd links `intersect` nubOrd associations
+  | (x:_) <- nubOrd strippedLinks `intersect` nubOrd associations
   = Just [iii|
       Link names and association names must be disjoint
       but currently "#{x}" is among both.
       |]
+  | strippedODMapping <- BM.mapMonotonic
+      stripNumericPeriod
```
jvoigtlaender: I'm now wondering which `mapMonotonic` or `mapMonotonicR` calls should actually be just `map` or `mapR` calls. After all, `stripNumericPeriod` is not really "strictly increasing" as the documentation of the functions using monotonicity demands.

patritzenfeld: Fair, I think mapMonotic was just used (before this pull request), because it is faster. But this function is not actually monotonically increasing.

PR #494, opened 2025-12-18
statistics of match-cd-od tasks

guard-who:
Average CPU and memory usage is also recorded for `match-cd-od` besides average time, but they are not available for `DifferentNames`, because I previously forgot adding them in the scripts.

### DIFF
--- a/example/src/Modelling/CdOd/DifferentNames/Config.hs
+++ b/example/src/Modelling/CdOd/DifferentNames/Config.hs
@@ -339,7 +339,7 @@ task2025_14 = DifferentNamesConfig {
 points: 0.15
 variant 1: concepts are printed in class diagrams
 share same instances as task2025_14
-average concept generation time per instance: 10~15 mins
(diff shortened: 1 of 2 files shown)


jvoigtlaender:
I guess I have a similar question for
> average concept generation time per instance: 10~15 mins

then.

Does the 10~15 mins, times 100 instances, mean that the whole injection run took about 20 hours? Or was some degree of concurrency used here as well?

guard-who:
> Does the 10~15 mins, times 100 instances, mean that the whole injection run took about 20 hours?

Yes, the "10~15 mins" refers to the time required for a single instance. Currently, the Python program injects concepts one by one. Concurrency is not yet implemented, but I will consider implementing it later. Also, the injection process was run on my computer, not on the cluster. There is no need to run it on the cluster.

It was so slow, because I used `gpt-5`, which is a thinking model, and the agent system allows maximal 5-round chats.

PR #526, opened 2026-01-05
use hasGivenCd flag to avoid conflicting showInputHelp

guard-who:
Close #525

PR #527, opened 2026-01-05
use hasGivenCd flag to provide targeted showInputHelp and mappingAdvice

### LINKED ISSUE #525: Avoid conflicting `showInputHelp` for `DifferentNames` task variants
Refer to [https://github.com/fmidue/llm-task-wrappers/pull/131#issuecomment-3709355803](https://github.com/fmidue/llm-task-wrappers/pull/131#issuecomment-3709355803).

jvoigtlaender:
See:
- https://github.com/fmidue/modelling-tasks/pull/526#issuecomment-3712122707
- https://github.com/fmidue/llm-task-wrappers/pull/131#issuecomment-3712135452
- https://git.uni-due.de/fmi/autotool-dev/-/issues/68#note_239220

### DIFF
--- a/src/Modelling/CdOd/DifferentNames.hs
+++ b/src/Modelling/CdOd/DifferentNames.hs
@@ -442,45 +442,45 @@ defaultDifferentNamesTaskText = [
 inputHelpText :: Bool -> Output
 inputHelpText hasGivenCd =
   if hasGivenCd then
-    Paragraph [
-      Translated $ translations $ do
-        english [iii|
-          State your answer by giving a mapping of
-          relationships in the CD to links in the OD.
-          \n
-          To state that x in the CD corresponds to 1. in the OD and
-          y in the CD corresponds to 2. in the OD, write the mapping as:
-          |]
-        german [iii|
-          Geben Sie Ihre Antwort als eine Zuordnung von
-          Beziehungen im CD zu Links im OD an.
-          \n
-          Um anzugeben, dass x im CD zu 1. im OD und y im CD
-          zu 2. im OD korrespondieren, schreiben Sie die Zuordnung als:
-          |],
-      Code . uniform . show $ mappingShow differentNamesInitial
-      ]
+  Paragraph [
+    Translated $ translations $ do
+      english [iii|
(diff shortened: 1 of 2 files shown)


jvoigtlaender:
Needs to be updated according to https://github.com/fmidue/modelling-tasks/pull/526#issuecomment-3712122707.

src/Modelling/CdOd/DifferentNames.hs
```diff
@@ -438,26 +439,49 @@ defaultDifferentNamesTaskText = [
...
+          |]
+        german [iii|
+          Geben Sie Ihre Antwort als eine Zuordnung von
+          Beziehungen im CD zu Links im OD an.
+          \n
+          Um anzugeben, dass x im CD zu 1. im OD und y im CD
+          zu 2. im OD korrespondieren, schreiben Sie die Zuordnung als:
+          |],
+      Code . uniform . show $ mappingShow differentNamesInitial
+      ]
```
jvoigtlaender: Use a non-standard indentation (not increasing the level by 2) in order to make the above diff much smaller.

src/Modelling/CdOd/DifferentNames.hs
```diff
@@ -356,29 +354,31 @@ toTaskText
...
+    toOutputCapable [inputHelpText hasGivenCd]
+  extra addText
   pure ()
+  where
+    hasGivenCd = Special GivenCd `elem` taskText
 
-mappingAdvice :: OutputCapable m => Bool -> LangM m
-mappingAdvice isCollapsed = collapsed isCollapsed (translations $ do
+mappingAdvice :: OutputCapable m => Bool -> Bool -> LangM m
+mappingAdvice isCollapsed hasGivenCd = collapsed isCollapsed (translations $ do
```
jvoigtlaender: I think we don't need two `Bool`s. I wanted `Folded True` in the CD-less cases. So it will always be the case (it seems) that `isCollapsed == not hasGivenCd`.

jvoigtlaender:
My comment `isCollapsed == not hasGivenCd` means that the call of `mappingAdvice` needs to be changed!

PR #539, opened 2026-01-12
Expose NameCdError answer texts

horriblename:
for the purpose of showing possible answers in the custom_form <https://git.uni-due.de/fmi/autotool-dev/-/issues/57#note_233475>

1. exposes functions to construct the translations of possible answers in NameCdError
2. expose a function to construct a version of defaultNameCdErrorTaskText that excludes the possible solutions since they will be shown in the custom_form already

### DIFF
--- a/src/Modelling/CdOd/NameCdError.hs
+++ b/src/Modelling/CdOd/NameCdError.hs
@@ -516,6 +516,9 @@ checkNameCdErrorInstance NameCdErrorInstance {..}
       . filter isRelevant
       $ annotatedRelationships classDiagram
 
+defaultNameCdErrorTaskText :: NameCdErrorTaskText
+defaultNameCdErrorTaskText = nameCdErrorTaskText True
+
 nameCdErrorTaskText :: Bool -> NameCdErrorTaskText
 nameCdErrorTaskText withAnswerChoices = concat [
  [
@@ -555,9 +558,6 @@ nameCdErrorTaskText withAnswerChoices = concat [
  ]]
  where
     optional cond xs = if cond then xs else []
-
-defaultNameCdErrorTaskText :: NameCdErrorTaskText
-defaultNameCdErrorTaskText = nameCdErrorTaskText True
 
 inputHelpText :: [Output]
 inputHelpText = [
--- a/src/Modelling/CdOd/NameCdError.hs
+++ b/src/Modelling/CdOd/NameCdError.hs
@@ -32,8 +32,10 @@ module Modelling.CdOd.NameCdError (
   nameCdErrorSyntax,
   nameCdErrorTask,
   nameCdErrorTaskText,
+  relevantRelationships,
   renameInstance,
   translateReason,
+  translateRelationship,
   ) where
 
 import qualified Modelling.CdOd.CdAndChanges.Transform as Changes (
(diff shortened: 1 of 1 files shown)


src/Modelling/CdOd/NameCdError.hs
```diff
@@ -236,10 +240,10 @@ isCustom = \case
   PreDefined {} -> False
 
 renderReason :: OutputCapable m => Bool -> Reason -> LangM m
-renderReason withDirections = translate . put . toTranslations withDirections
+renderReason withDirections = translate . put . translateReason withDirections
 
-toTranslations :: Bool -> Reason -> Map Language String
-toTranslations withDirections = \case
+translateReason :: Bool -> Reason -> Map Language String
+translateReason withDirections = \case
```
jvoigtlaender, line 246: Let's not rename functions without need.

src/Modelling/CdOd/NameCdError.hs
```diff
@@ -416,19 +420,29 @@ toTaskSpecificText path task@NameCdErrorInstance {..} = \case
...
-          phraseRelationship' Annotation {..} = phrase
-            (referenceUsing annotation)
+      let phraseRelationship' = translateRelationship
+            cdDrawSettings
             byName
-            (printNavigations cdDrawSettings)
-            annotated
       enumerateM (text . show)
-        $ map (second phraseRelationship')
+        $ map (second (translate . put . phraseRelationship'))
```
jvoigtlaender, line 427: Why is the `translate . put` part necessary that wasn't there before?

src/Modelling/CdOd/NameCdError.hs
```diff
@@ -523,15 +542,22 @@ defaultNameCdErrorTaskText = [
...
     english [i|The class diagram ...|]
     german [i|Das Klassendiagramm ...|],
   Paragraph $ singleton $ Special ReasonsList
-  ]
+ ]]
+ where
+    optional cond xs = if cond then xs else []
+
+defaultNameCdErrorTaskText :: NameCdErrorTaskText
+defaultNameCdErrorTaskText = nameCdErrorTaskText True
```
jvoigtlaender: Put `defaultNameCdErrorTaskText` above `nameCdErrorTaskText`, to minimze diff.

src/Modelling/CdOd/NameCdError.hs
horriblename: `phraseRelationship'` in the old diff returns a `State (Map Language String)` while the new one returns a `Map Language String` (which in turn is because I don't think it makes sense for the exposed function `translateRelationship` to return a `State`)

src/Modelling/CdOd/NameCdError.hs
horriblename: since I'm exposing this function, I think it's better to give it a less ambiguous name

src/Modelling/CdOd/NameCdError.hs
jvoigtlaender: I don't see how previously it returned a `State ...`. It called `translate`, so must have returned a `GenericLangM language m ()`. Maybe it's still better to return a `Map ...`, though.

PR #547, opened 2026-01-13
add 2025 task configs of `Reach` and `Deadlock`

guard-who:


guard-who:
For `task2025_27 `:
From:
image
to:
image

I tried to generate some instances. In the above example, I think most of properties mentioned in #226 have been considered. but the property "Anfangs- und Endnetz unterscheiden sich nur durch die Tokenzahl an genau einer Stelle (oder ein Parameter k dafür)." has been not enforced yet. And `absentTransitionsRequirement` is set `0` to make some differences to `task2025_28`.  

The solution should be `t4 -> t2 -> t1 -> t4 -> t2 -> t1 -> t2 -> t3`. Follow your analysis in the linked issue, then:
- Initially, only `t1` can increase the number of tokens, so we need to trigger `t1` at least two times.
- Before `t1` is activatable, `t4, t2` should be triggered in sequence to switch tokens.
- Then, we can repeat the cycle two times and get enough tokens in the net.
- Finally, we should trigger `t2, t3` to move tokens in the correct places.

guard-who:
In the newest commits, via changing `requireCycleLengthsAny` and `maxPlacesChanged`, now I can't find any instance which violate the properties mentioned in the linked issue for `Reach` tasks.

In my vision, the current configuration should always allow a solution that can be obtained by decomposing the sequence into two parts (e.g., 8 (transitionLength) = 4 + 4 for `task2025_27`, 12 = 6 + 6 for `task2025_28`) and then obtaining the complete result by repeating the subsequence (about [https://github.com/fmidue/modelling-tasks/issues/226#issuecomment-2517875808](https://github.com/fmidue/modelling-tasks/issues/226#issuecomment-2517875808)). However, for `task2025_28`, I am not sure about if a subsequence with the length 5 is possible, then we have a solution with two this subsequence and two other transitions. Smaller decomposition should be constrained by `transitionCoverageRequirement`.

Note: I only check about 10 instances, so I am not sure about if they are totally guaranteed by current implementation.

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -123,3 +124,87 @@ task2024_60 = ReachConfig {
...
+    numTransitions = 4,
+    capacity = Unbounded,
+    graphLayouts = [Circo],
+    maxTransitionLength = 8,
+    minTransitionLength = 8,
+    arrowDensityConstraints = ArrowDensityConstraints {
+      incomingArrowsPerTransition = (2, Just 2),
+      outgoingArrowsPerTransition = (2, Just 3),
+      incomingArrowsPerPlace = (0, Nothing),
+      outgoingArrowsPerPlace = (0, Nothing),
```
jvoigtlaender: Can you set `incomingArrowsPerPlace = (2, Just 3)` and `outgoingArrowsPerPlace = (1, Just 3)`?

jvoigtlaender: Actually no, just make both to be `(1, Nothing)`.

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -123,3 +124,87 @@ task2024_60 = ReachConfig {
...
+      totalArrowsFromPlacesToTransitions = (8, Just 8),
+      totalArrowsFromTransitionsToPlaces = (8, Just 12)
```
jvoigtlaender: And herem change the `Just 12` to `Just 10`?

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -123,3 +124,87 @@ task2024_60 = ReachConfig {
...
+    },
+  maxPrintedSolutions = 10,
+  rejectLongerThan = Just 8,
+  showLengthHint = False,
+  showMinLengthHint = True,
+  showTargetNet = True,
+  showPlaceNamesInNet = False,
+  filterConfig = FilterConfig {
+    rejectGroupedRepeats = True,
+    repetitiveSubsequenceThreshold = Just 3,
```
jvoigtlaender: Set this to `Just 4`. The setting with `3` would have outlawed the example in the referenced issue.

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -123,3 +124,87 @@ task2024_60 = ReachConfig {
...
+    spaceballsPrefixThreshold = Just 4,
+    forbiddenCycleLengths = [],
+    requireCycleLengthsAny = [4],
+    solutionSetLimit = Just 15,
+    requireSolutionsArePermutations = True,
+    absentTransitionsRequirement = 0,
```
jvoigtlaender, line 174: Set this to `1`? That was the case in the referenced example.

jvoigtlaender: >  And `absentTransitionsRequirement` is set `0` to make some differences to `task2025_28`.  

Okay.

jvoigtlaender:
What do you mean by
> but the property "Anfangs- und Endnetz unterscheiden sich nur durch die Tokenzahl an genau einer Stelle (oder ein Parameter k dafür)." has been not enforced yet

?

That you didn't set the config option to enforce that, or that you did set the config but it didn't have the impact it should have had?

jvoigtlaender:
I'm also not sure what you mean by this:
> In the newest commits, via changing `requireCycleLengthsAny` and `maxPlacesChanged`, now I can't find any instance which violate the properties mentioned in the linked issue for `Reach` tasks.

jvoigtlaender:
> However, for `task2025_27`, I am not sure ...

Do you mean `task2025_28` here?

guard-who:
> Do you mean task2025_28 here?

Yes, sorry for the mistake.

jvoigtlaender:
For `task2025_27`, maybe we shouldn't require a cycle length. With 8 = 4 + 4, and there being 4 transitions and `transitionCoverageRequirement = 1 % 1`, it would mean that there must be a solution with some permutation of `t1...t4` and that permutation repeated twice. That's maybe too specific.

jvoigtlaender:
Maybe we should instead actually forbid cycle length `4` in that case.

guard-who:
> What do you mean by
> 
> > but the property "Anfangs- und Endnetz unterscheiden sich nur durch die Tokenzahl an genau einer Stelle (oder ein Parameter k dafür)." has been not enforced yet
> 
> ?
> 
> That you didn't set the config option to enforce that, or that you did set the config but it didn't have the impact it should have had?

I didn't set it at that time, because I was not aware that `maxPlacesChanged` is used for this property, but it was fixed in the new commits.

> I'm also not sure what you mean by this:
> 
> > In the newest commits, via changing `requireCycleLengthsAny` and `maxPlacesChanged`, now I can't find any instance which violate the properties mentioned in the linked issue for `Reach` tasks.

The example I showed above violated some properties mentioned in the linked issue #226  (mainly because of `maxPlacesChanged`), but after new changes, generated instances seems more matched with the expected examples in the linked issue.

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -123,3 +124,87 @@ task2024_60 = ReachConfig {
...
+    numTransitions = 6,
+    capacity = Unbounded,
+    graphLayouts = [Circo],
+    maxTransitionLength = 12,
+    minTransitionLength = 12,
+    arrowDensityConstraints = ArrowDensityConstraints {
+      incomingArrowsPerTransition = (2, Just 2),
+      outgoingArrowsPerTransition = (2, Just 3),
+      incomingArrowsPerPlace = (0, Nothing),
+      outgoingArrowsPerPlace = (0, Nothing),
```
jvoigtlaender: Here you could set `incomingArrowsPerPlace = (2, Just 3)` and `outgoingArrowsPerPlace = (1, Just 3)`.

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -135,12 +135,12 @@ task2025_27 = ReachConfig {
     maxTransitionLength = 8,
     minTransitionLength = 8,
     arrowDensityConstraints = ArrowDensityConstraints {
-      incomingArrowsPerTransition = (2, Just 2),
+      incomingArrowsPerTransition = (2, Just 3),
```
jvoigtlaender: Where does this change suddenly come from?

jvoigtlaender:
> The example I showed above violated some properties

To clarify again: Is this about the config options not being respected by the generator somehow?

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -123,3 +124,87 @@ task2024_60 = ReachConfig {
...
+    graphLayouts = [Circo],
+    maxTransitionLength = 12,
+    minTransitionLength = 12,
+    arrowDensityConstraints = ArrowDensityConstraints {
+      incomingArrowsPerTransition = (2, Just 2),
+      outgoingArrowsPerTransition = (2, Just 3),
+      incomingArrowsPerPlace = (2, Just 3),
+      outgoingArrowsPerPlace = (1, Just 3),
+      totalArrowsFromPlacesToTransitions = (12, Just 12),
+      totalArrowsFromTransitionsToPlaces = (12, Just 18)
```
jvoigtlaender: Narrow this to `(14, Just 16)`.

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -123,3 +124,87 @@ task2024_60 = ReachConfig {
...
+      },
+    maxPlacesChanged = 1,
+    transitionBehaviorConstraints = TransitionBehaviorConstraints {
+      allowedTokenChanges = Just GT,
+      areNonPreserving = Just 1
```
jvoigtlaender: I suggest `Just 2` here. That's what the larger `Reach.Reach` example in the linked issue had.

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -123,3 +124,87 @@ task2024_60 = ReachConfig {
...
+  filterConfig = FilterConfig {
+    rejectGroupedRepeats = True,
+    repetitiveSubsequenceThreshold = Just 3,
+    spaceballsPrefixThreshold = Just 4,
+    forbiddenCycleLengths = [],
+    requireCycleLengthsAny = [6],
+    solutionSetLimit = Just 15,
+    requireSolutionsArePermutations = True,
+    absentTransitionsRequirement = 1,
+    transitionCoverageRequirement = 4 % 5
```
jvoigtlaender: This means that five transitions must be present. The example in the referenced issue used only four of its six transitions. So I suggest a lower bound here. I would then accept both cycle lengths `4` and `6`.

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -123,3 +124,87 @@ task2024_60 = ReachConfig {
...
+  showTargetNet = True,
+  showPlaceNamesInNet = False,
+  filterConfig = FilterConfig {
+    rejectGroupedRepeats = True,
+    repetitiveSubsequenceThreshold = Just 4,
+    spaceballsPrefixThreshold = Just 4,
+    forbiddenCycleLengths = [4],
+    requireCycleLengthsAny = [],
+    solutionSetLimit = Just 15,
+    requireSolutionsArePermutations = True,
```
jvoigtlaender: Maybe here we can set `requireSolutionsArePermutations = False`, i.e., permit more diverse solution variety.

guard-who:
> > The example I showed above violated some properties
> 
> To clarify again: Is this about the config options not being respected by the generator somehow?

No, `maxPlacesChanged` was set 4 at that time and I didn't set anything about `forbiddenCycleLengths` and `requireCycleLengthsAny`, so I easily get the example, which should be allowed for the config before changes, but I currently can't easily get any similar instances after changes.

example/src/Modelling/PetriNet/PetriReach/Config.hs
```diff
@@ -201,10 +201,10 @@ task2025_28 = ReachConfig {
     repetitiveSubsequenceThreshold = Just 3,
     spaceballsPrefixThreshold = Just 4,
     forbiddenCycleLengths = [],
-    requireCycleLengthsAny = [6],
+    requireCycleLengthsAny = [],
```
jvoigtlaender: What happened to this?
>  I would then accept both cycle lengths 4 and 6.

guard-who: Should I set it? Because you said "I would then accept both cycle lengths 4 and 6." Should `requireCycleLengthsAny = [6]` only require that there is at least one solution with the cycle length 6 in the solution set? So I think it is not required now, unless we want each task should have a solution with the cycle length 6.

jvoigtlaender: Setting `requireCycleLengthsAny = [4,6]` will have the desired effect: There must be a cycle, it is allowed to have either length 4 or length 6.

example/src/Modelling/PetriNet/PetriDeadlock/Config.hs
```diff
@@ -127,28 +127,28 @@ task2025_29 = DeadlockConfig {
...
     },
   maxPrintedSolutions = 10,
   rejectLongerThan = Just 7,
   showLengthHint = False,
   showMinLengthHint = True,
   showPlaceNamesInNet = False,
-  requireFusableInputNodes = Nothing,
-  requireFusableOutputNodes = Nothing,
+  requireFusableInputNodes = Just 1,
+  requireFusableOutputNodes = Just 1,
```
jvoigtlaender: With only four transitions and four places, it's maybe a bit much to ask that two places and two transitions must play special roles.

example/src/Modelling/PetriNet/PetriDeadlock/Config.hs
```diff
@@ -154,40 +154,40 @@ task2025_29 = DeadlockConfig {
...
+      spaceballsPrefixThreshold = Just 4,
+      forbiddenCycleLengths = [],
+      requireCycleLengthsAny = [],
+      solutionSetLimit = Just 15,
+      requireSolutionsArePermutations = False,
+      absentTransitionsRequirement = 1,
+      transitionCoverageRequirement = 4 % 6
+      }
     }
-  }
```
guard-who: It is still under running. This task generally requires much more time than other three to find an instance.

example/src/Modelling/PetriNet/PetriDeadlock/Config.hs
```diff
@@ -110,3 +111,83 @@ task2024_61 = DeadlockConfig {
...
+  showPlaceNamesInNet = False,
+  requireFusableInputNodes = Just 1,
+  requireFusableOutputNodes = Just 1,
+  filterConfig = FilterConfig {
+    rejectGroupedRepeats = True,
+    repetitiveSubsequenceThreshold = Just 4,
+    spaceballsPrefixThreshold = Just 4,
+    forbiddenCycleLengths = [],
+    requireCycleLengthsAny = [],
+    solutionSetLimit = Just 15,
```
jvoigtlaender, line 206: Maybe you should set this to `Nothing` to give the system a better chance of accepting a generated net.

example/src/Modelling/PetriNet/PetriDeadlock/Config.hs
```diff
@@ -110,3 +111,83 @@ task2024_61 = DeadlockConfig {
...
+  requireFusableInputNodes = Just 1,
+  requireFusableOutputNodes = Nothing,
+  filterConfig = FilterConfig {
+    rejectGroupedRepeats = True,
+    repetitiveSubsequenceThreshold = Just 3,
+    spaceballsPrefixThreshold = Just 4,
+    forbiddenCycleLengths = [],
+    requireCycleLengthsAny = [],
+    solutionSetLimit = Just 15,
+    requireSolutionsArePermutations = False,
```
jvoigtlaender, line 158: Maybe this here can be set to `True`?

example/src/Modelling/PetriNet/PetriDeadlock/Config.hs
```diff
@@ -110,3 +111,83 @@ task2024_61 = DeadlockConfig {
...
+    absentTransitionsRequirement = 0,
+    transitionCoverageRequirement = 3 % 4
```
jvoigtlaender: And here maybe all transitions should be required.

example/src/Modelling/PetriNet/PetriDeadlock/Config.hs
guard-who: It made the generation much slower, so I prefer to keep it if it doesn't have a large impact on the solution strategy for `deadlock` task under this config (I think so).

guard-who: > It made the generation much slower

For example, I have just tried to change `transitionCoverageRequirement = 1 % 1` and `requireSolutionsArePermutations = True`, then I can't get any instance within 10 mins, but I can get one instance within 10 seconds if `transitionCoverageRequirement = 1 % 1` and `requireSolutionsArePermutations = False`.

example/src/Modelling/PetriNet/PetriDeadlock/Config.hs
```diff
@@ -110,3 +111,83 @@ task2024_61 = DeadlockConfig {
...
+  transitionBehaviorConstraints = TransitionBehaviorConstraints {
+    allowedTokenChanges = Nothing,
+    areNonPreserving = Nothing
+    },
+  arrowDensityConstraints = ArrowDensityConstraints {
+    incomingArrowsPerTransition = (1, Just 2),
+    outgoingArrowsPerTransition = (1, Just 2),
+    incomingArrowsPerPlace = (1, Just 3),
+    outgoingArrowsPerPlace = (1, Just 2),
+    totalArrowsFromPlacesToTransitions = (9, Just 9),
```
jvoigtlaender: I imagine you should widen this range.

example/src/Modelling/PetriNet/PetriDeadlock/Config.hs
```diff
@@ -110,3 +111,83 @@ task2024_61 = DeadlockConfig {
...
+  filterConfig = FilterConfig {
+    rejectGroupedRepeats = True,
+    repetitiveSubsequenceThreshold = Just 4,
+    spaceballsPrefixThreshold = Just 4,
+    forbiddenCycleLengths = [],
+    requireCycleLengthsAny = [],
+    solutionSetLimit = Just 15,
+    requireSolutionsArePermutations = False,
+    absentTransitionsRequirement = 1,
+    transitionCoverageRequirement = 4 % 6
```
jvoigtlaender: `4 % 6` is actually a "weird" expression; why not `2 % 3`?

example/src/Modelling/PetriNet/PetriDeadlock/Config.hs
```diff
@@ -110,3 +111,83 @@ task2024_61 = DeadlockConfig {
...
+    totalArrowsFromPlacesToTransitions = (9, Just 9),
+    totalArrowsFromTransitionsToPlaces = (10, Just 14)
+    },
+  maxPrintedSolutions = 10,
+  rejectLongerThan = Just 14,
+  showLengthHint = False,
+  showMinLengthHint = True,
+  showPlaceNamesInNet = False,
+  requireFusableInputNodes = Just 1,
+  requireFusableOutputNodes = Just 1,
```
jvoigtlaender: The concrete old task instance had more than two "fusion situations". Actually, putting a "too low" number here might conflict with the relatively low `totalArrows` bounds. If there are few arrows in the net, then it becomes more likely that fusable situations appear (because more often is there only one transition consuming from a given place etc., simply because there are few arrows put in the net). But then if for example the settings say that only `2` fusable situations are allowed to (and have to) exist, then that might not be satisfiable.

jvoigtlaender:
Note:
```
modelling-tasks-examples   > Failures:
modelling-tasks-examples   > 
modelling-tasks-examples   >   test/ExamplesSpec.hs:510:38: 
modelling-tasks-examples   >   1) Examples, check config, 2025, task27
modelling-tasks-examples   >        expected: Nothing
modelling-tasks-examples   >         but got: Just "with allowedTokenChanges = Just GT and areNonPreserving = Just 1, lower bound difference between totalArrowsFromPlacesToTransitions and totalArrowsFromTransitionsToPlaces cannot be just 0"
modelling-tasks-examples   > 
modelling-tasks-examples   >   To rerun use: --match "/Examples/check config/2025/task27/" --seed 727989311
modelling-tasks-examples   > 
modelling-tasks-examples   >   test/ExamplesSpec.hs:512:38: 
modelling-tasks-examples   >   2) Examples, check config, 2025, task28
modelling-tasks-examples   >        expected: Nothing
modelling-tasks-examples   >         but got: Just "with areNonPreserving = Just 2 and at most 1 token increase per transition, overall at most 2 token increase is possible, but upper bound difference between totalArrowsFromPlacesToTransitions and totalArrowsFromTransitionsToPlaces is 4"
modelling-tasks-examples   > 
modelling-tasks-examples   >   To rerun use: --match "/Examples/check config/2025/task28/" --seed 727989311
```

This could mean that some checks introduced in #553 / 8e8acde849db6ef7b7bc827b98b562ba6d94b41a are too strict, or that something should be changed in the configs here.

example/src/Modelling/PetriNet/PetriDeadlock/Config.hs
guard-who: Perhaps, is `3 % 4` better? I didn't notice that the config has `8` transition instead of `6`. Or exactly speaking, I noticed it, but I didn't think here, just moved the one in `task2025_28` here.

guard-who:
`task2025_30` now can generate one instance within 10 minutes, but it changes the solution strategy (compared with the one in the linked issue) a bit. I think the strategy is still relatively fixed for students. Considering the following example:

image

`t1` and `t7` require two incoming transitions, but only the place `t6` pointed to has no other transitions which can be activated, so tokens should be moved to the place. Then just follow the fusion strategy, students can easily find the correct solution. Honestly, I think it is a little bit too easy 😊. What do you think of?

jvoigtlaender:
I think that's a good task instance.

guard-who:
Okay, then I start generating enough instances for these tasks.

guard-who:
All instances for Task 27 - 30 are now available on Autotool.

PR #560, opened 2026-01-19
add ToDoc and Reader instances needed for current generation

### LINKED ISSUE #559: Provide more `ToDoc` and `Reader` instances
[ReachInstance](https://github.com/fmidue/modelling-tasks/blob/7489112bdb3017f0196e4dcacce80403c7ee79af/src/Modelling/PetriNet/Reach/Reach.hs#L471C1-L490C39), [DeadlockInstance](https://github.com/fmidue/modelling-tasks/blob/7489112bdb3017f0196e4dcacce80403c7ee79af/src/Modelling/PetriNet/Reach/Deadlock.hs#L214-L232), and some others of those task instances not appeared in the repo `llm-task-wrappers` have also not derived `ToDoc` and `Reader` instances. We need to provide them here instead of in `autotool-dev` to support directly generating Autotool-readable instances on the cluster for the task deployment on Autotool.

patritzenfeld:
- ReachInstance
- DeadlockInstance

as requested in #559

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -41,7 +41,6 @@ dependencies:
   - autolib-todoc
   - autolib-util
   - autotool-capabilities
-  - autotool-capabilities-io-instances
   - base
   - bimap
   - bifunctors
--- a/package.yaml
+++ b/package.yaml
@@ -41,6 +41,7 @@ dependencies:
   - autolib-todoc
   - autolib-util
   - autotool-capabilities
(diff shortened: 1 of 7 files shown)


patritzenfeld:
Pushed something by accident, will remove that first

patritzenfeld:
This should now handle #559 completely.

Issue #562, opened 2026-01-20
Hovering doesn't work properly on Petri net places with high token count


jvoigtlaender:
Note this:

<img width="448" height="418" alt="Image" src="https://github.com/user-attachments/assets/278e00f7-e8b9-4e20-90c5-bf2965786da3" />

and this:

<img width="353" height="330" alt="Image" src="https://github.com/user-attachments/assets/8c1c804f-5235-46d0-9e36-619744be5fe8" />

vs. this:

<img width="376" height="340" alt="Image" src="https://github.com/user-attachments/assets/4688bbd2-0fc6-477b-bba2-ae2556f12d5a" />

nimec01:
The issue here is that the token count is not displayed, right?

Could you provide a concrete task instance where this happens?

jvoigtlaender:
The problem is that the "7 bullets" inscription does not react to hovering at all, neither actively nor passively. In contrast to how "several bullets" displayed directly (without number printing) react.

Let's take this:

<img width="459" height="452" alt="Image" src="https://github.com/user-attachments/assets/402fa207-585c-4d04-9406-963815ce2a59" />

If I move the mouse pointer to the circle border, I get this:

<img width="513" height="453" alt="Image" src="https://github.com/user-attachments/assets/f3a28709-254a-41b1-a327-289a5536aebe" />

And when I move the mouse pointer to the bullets themselves, again:

<img width="478" height="410" alt="Image" src="https://github.com/user-attachments/assets/f17907b3-1da8-44f9-ab24-d88993e3ecf6" />

Now consider this:

<img width="361" height="367" alt="Image" src="https://github.com/user-attachments/assets/56c3791a-a06a-4474-af1a-f5fabaa659c6" />

If I move the mouse pointer to the circle border, I get this:

<img width="377" height="373" alt="Image" src="https://github.com/user-attachments/assets/2c4af72e-d010-4340-87ec-6d0a1cf28cbb" />

And when I move the mouse pointer to the number or the bullet, I get this:

<img width="337" height="331" alt="Image" src="https://github.com/user-attachments/assets/c9058af0-7624-4abb-8293-42cbf22d4717" />

An example instance where this happens (in the goal net) is this:
```
Right (ReachInstance 
  { netGoal = (NetGoal 
      { drawUsing = Circo
      , petriNet = (Net 
          { places = (mkSet [ s1, s2, s3, s4 ])
          , transitions = (mkSet [ t1, t2, t3, t4 ])
[...]

Issue #260, opened 2024-06-24
emitting label info into the `svg`s


jvoigtlaender:
For the external use in the BA thesis, the student needs to know which labels from the algebraic data correspond to which parts of a generated image.

For example, he has obtained this image: 
![petrid1275ea9040a8a212a4e2efbae9a2dcdb844d4a840c10b53f32cf0a09f552ed10010](https://github.com/jvoigtlaender/modelling-tasks-internal/assets/5853832/aff4bda6-6608-47b9-88e4-7f0d684a5ca3)
and now wants to be able to have the user click into the image and know which transition or place was clicked on. So he needs to know what the nodes "mean".

He has identified that the labels are currently rendered as svg paths. For example, the `7` in the picture is this piece in the svg file:
```svg
<g fill="rgb(0,0,0)" fill-opacity="0.0" stroke="rgb(0,0,0)" stroke-linecap="butt" stroke-linejoin="miter" stroke-miterlimit="10.0" stroke-opacity="1.0" stroke-width="0.36646144825564353"><path class="rect" d="M 148.0823,186.9906 l -0.0000,-14.6585 h -14.6585 l -0.0000,14.6585 Z" fill="none" fill-opacity="0.0"/><path class="" d="M 138.5320,183.8121 ZM 140.1560,178.7013 c -0.3057,-0.0000 -0.6209,0.0669 -0.8311 0.8311c -0.1051,0.0124 -0.2016,0.0010 -0.3057 -0.0287c 0.0764,-0.4968 0.1815,-1.0699 0.1911 -1.5285c 0.0000,-0.0287 0.0096,-0.0287 0.0382 -0.0287c 0.1720,0.0382 0.1528,0.1337 0.4776 0.1337h 1.9679 c 0.3630,-0.0000 0.5732,-0.0573 0.7356 -0.1051l 0.1528,0.1146 c -1.0699,2.6080 -1.5953,4.1937 -2.1399 5.8177l -0.6496,0.0287 l -0.0764,-0.0669 c 0.6496,-1.4520 1.4043,-3.3149 2.1494 -5.1681h -1.7100 Z" fill-opacity="1.0" stroke-width="0.0"/></g>
```
There's nothing obvious in there that signals `7`.

Is it possible to emit such information into the `svg` file, via hidden attributes or some such?

jvoigtlaender:
[...]

PR #572, opened 2026-01-28
Hoveringanpassungen bei Aufgaben mit Petrinetzanzeige

### LINKED ISSUE #229: Adaption Hovering bei Aufgaben mit Petrinetzanzeige
Zum Beispiel bei `SelectPetri` wird der Hinweis, dass über Diagrammelemente gefahren werden kann, um "Zusammengehörigkeit" von Pfeilen etc. herauszufinden, nicht ausgegeben.

Ich bin mir nicht sicher, ob dies Absicht ist, also bei diesem (und weiteren?) Aufgabentypen das bewusst weggelassen wurde.

Argumentiert werden könnte, dass der Hinweis konkret bei `SelectPetri` nicht so relevant ist, da es keine Gewichte im Petrinetz gibt und auch eventuelle Stellen- und Transitionsnamen nicht wirklich wichtig sind (weil sie in der Lösungseingabe nicht gebraucht werden), somit das Hovern nicht besonders wichtig ist. Zumal außerdem in den angezeigten ADs sowieso kein Highlighting-per-Hovering möglich ist.

Allerdings könnte dann auch argumentiert werden, dass in dem Fall das Hovering auch für die Petrinetze sogar ausgeschaltet werden sollte?

nimec01:
Closes #229


src/Modelling/ActivityDiagram/SelectPetri.hs
```diff
@@ -205,12 +206,14 @@ checkSelectPetriConfig' SelectPetriConfig {
...
     petriLayout,
+    petriSvgHighlighting,
     numberOfWrongAnswers,
     numberOfModifications,
     auxiliaryPetriNodeAbsent,
     presenceOfSinkTransitionsForFinals,
     withActivityFinalInForkBlocks
   } = validateSelectPetriSpecific numberOfWrongAnswers numberOfModifications
+    <|> if petriSvgHighlighting then Nothing else Just "petriSvgHighlighting must be enabled for this task."
     <|> validatePetriConfig
```
jvoigtlaender, line 217: Careful (also elsewhere): This is almost certainly parsed as
```
if petriSvgHighlighting then Nothing else
 (Just "petriSvgHighlighting must be enabled for this task." <|> validatePetriConfig ... )
```
which is not what you want.

jvoigtlaender: Can you add two rules to `.hlint.yaml` that will catch this by proposing rewrites
```
Nothing <|> x  -->  x
Just x <|> y --> Just x
```
The rule format is like this: https://github.com/ndmitchell/hlint/blob/514ad8fae7774f1d4677566e95028cf7531bb5c2/data/hlint.yaml#L652-L653

nimec01:
Funktioniert `withSvgHighlighting = False` aktuell wirklich, wie es in https://github.com/fmidue/modelling-tasks/issues/229#issuecomment-2539767910 vermutet wird?

Also führt das Ausschalten auch tatsächlich dazu, dass dies nicht mehr möglich ist?

Wenn ich diese Änderungen hier nämlich teste, führt dies bei mir nicht dazu, dass beim Hovern kein Highlighten mehr stattfindet. Beim Erstellen der SVGs wird aber der korrekte Wert `withSvgHighlighting = False` übergeben. Wie Marcellus in einem anderen Kommentar schon angemerkt hatte, ist das SVG Highlighting generell eingeschaltet. Um dies auszuschalten, dürften im SVG, meinem Verständnis nach, keine Klassen wie "node", "edge", ... zugewiesen werden. Die Klassen werden aber anscheinend trotz `withSvgHighlighting = False` immernoch hinzugefügt.

jvoigtlaender:
Well this comment: https://github.com/fmidue/modelling-tasks/issues/260#issuecomment-2194008810 stated that yes, stuff is still annotated when this setting is used, but the annotations are somehow problematically structured then (for reasons beneficial elsewhere) and that
> destroys the structure for highlighting

So, the highlighting should be effectively (though "accidentally") turned off, then.

I guess the first thing to check now is whether flipping that config bit makes any difference for the generated `svg` at all. (It should have.)

jvoigtlaender:
Presumably the changes to `src/Modelling/PetriNet/Diagram.hs` in https://github.com/fmidue/modelling-tasks/commit/e2d0933f5ac48dec2f21edbae0e0348807065b1f are what "destroyed the structure".

That is, the `svgClass $ ' ' : l` in the last line here:

https://github.com/fmidue/modelling-tasks/blob/ded1ef65d294ef07e1ce80e2590483a8501f3850/src/Modelling/PetriNet/Diagram.hs#L251-L253

is what should be different/relevant between `svg`s intended for highlighting and `svg`s for which highlighting would/should not work.

nimec01:
> I guess the first thing to check now is whether flipping that config bit makes any difference for the generated `svg` at all. (It should have.)

The generated SVGs seem to be different. There are additional classes for the elements in the "without highlight" SVG (like `class="rect"` vs `class="rect t4`). But this does not make any difference for me in terms of highlighting. 

[Interestingly, the class names also differ for me when I generate them through Autotool or manually in `ghci` (`class="rect t4` vs `class="rect Transition 4`). But that does also not make any difference.]

nimec01:
Could it be that the “hack” no longer works? I generally mislike the idea to implement features by abusing some kind of "bug". 

It probably shouldn't be too hard to just add a class to the (root) SVG element in order to disable highlighting (also updating the CSS rules accordingly).

Also: attaching arbitrary data to any HTML (or SVG) element is pretty easy. You can just use `data-*` attributes. Implementing label annotations in this way should not affect the rendering/highlighting of the SVG at all.

That way it would be easier to separate the two concerns.

jvoigtlaender:
Okay. I think the best way forward, then, is:
- Make the `additionalLabel` printing in `drawNode` not dependent on the `withSvgHighlighting` parameter, but instead on an additional parameter `withoutAdditionalLabels` of the `drawNode`/`drawGraph` functions which is always set to `True` in this codebase.
- Use the `withSvgHighlighting` parameter in `drawNode` to instead turn on/off the `svgClass`-annotating overall.
- Then proceed in the spirit of #229.

jvoigtlaender:
(That was written without seeing your comment. Maybe your approach is ultimately the conceptually better one. But there are other constraints here, like what the additional labels were originally meant for, and also a desire to not otherwise change the css now. I'd opt for the minimal immediate change that will have the desired impact.)

nimec01:
It seems like my concept isn't that easy to implement here. I will elaborate more in #562.

jvoigtlaender:
You introduced a new parameter into the lecturer-facing config data types. That is not what I had in mind above:
> - Make the `additionalLabel` printing in `drawNode` not dependent on the `withSvgHighlighting` parameter, but instead on an additional parameter `withoutAdditionalLabels` of the `drawNode`/`drawGraph` functions which is always set to `True` in this codebase.

That was purely about having an additional parameter in the `drawNode`/`drawGraph` functions, and setting that parameter *always* (not only when the lecturer said or did not say something in the config) to `True` in this codebase.

(Someone outside this codebase, but using it as a library, might call these functions with that parameter set to `False`, though. That's the design I had in mind.)

nimec01:
> You introduced a new parameter into the lecturer-facing config data types. That is not what I had in mind above:
> 
> > * Make the `additionalLabel` printing in `drawNode` not dependent on the `withSvgHighlighting` parameter, but instead on an additional parameter `withoutAdditionalLabels` of the `drawNode`/`drawGraph` functions which is always set to `True` in this codebase.
> 
> That was purely about having an additional parameter in the `drawNode`/`drawGraph` functions, and setting that parameter _always_ (not only when the lecturer said or did not say something in the config) to `True` in this codebase.
> 
> (Someone outside this codebase, but using it as a library, might call these functions with that parameter set to `False`, though. That's the design I had in mind.)

Got it. However, always setting that variable to `True` i.e. enforcing label annotations wouldn't allow us to use SVG highlighting anymore (as described [here](https://github.com/fmidue/modelling-tasks/issues/562#issuecomment-3836702494) and [here](https://github.com/fmidue/modelling-tasks/issues/562#issuecomment-3837178401))

jvoigtlaender:
How does setting a parameter `withoutAdditionalLabels` to `True` amount to "enforcing label annotations"? :smile:

nimec01:
I think I'm done here. In order for the `no-highlights` SVG class to work, we need to make a tiny change to the CSS. I will open a new pull request there when this PR is approved.

src/Modelling/PetriNet/Diagram.hs
```diff
@@ -122,13 +127,15 @@ drawNet
   -- ^ the graph definition
   -> DrawSettings
   -- ^ how to draw the graph
+  -> Bool
+  -- ^ whether labels should be annotated
```
jvoigtlaender: The documentation should probably be more explicit about what labels/annotations this is about. 

A reader of the code might easily think that this is about something like labels going to be visible in the actual diagram display on screen.

src/Modelling/ActivityDiagram/MatchPetri.hs
```diff
@@ -379,6 +382,8 @@ matchPetriTask path task = do
         |]
     pure ()
 
+  hoveringInformation True
```
jvoigtlaender: Wouldn't it be more future-proof to "take this `Bool` from the config"?

src/Modelling/ActivityDiagram/SelectPetri.hs
```diff
@@ -205,12 +206,14 @@ checkSelectPetriConfig' SelectPetriConfig {
...
     maxInstances,
     petriLayout,
+    petriSvgHighlighting,
     numberOfWrongAnswers,
     numberOfModifications,
     auxiliaryPetriNodeAbsent,
     presenceOfSinkTransitionsForFinals,
     withActivityFinalInForkBlocks
   } = validateSelectPetriSpecific numberOfWrongAnswers numberOfModifications
+    <|> (if petriSvgHighlighting then Nothing else Just "petriSvgHighlighting must be enabled for this task.")
```
jvoigtlaender: The check here is the opposite of the one I asked for in https://github.com/fmidue/modelling-tasks/issues/229#issuecomment-2601667658.

> - bei `SelectPetri` einen Hovering-Hinweis neu ausgeben sofern `petriSvgHighlighting = True` gesetzt wäre, aber per Konfigurations-Checker zunächst erzwingen, dass immer `petriSvgHighlighting = False` gesetzt ist

src/Modelling/ActivityDiagram/SelectPetri.hs
```diff
@@ -400,6 +403,8 @@ Bitte geben Sie Ihre Antwort als Zahl an, welche das passende Petrinetz repräse
       german  [i|bedeuten, dass Petrinetz 2 das passende Petrinetz ist.|]
     pure ()
 
+  hoveringInformation True
```
jvoigtlaender: Again, this should be influenced by the config option, as said earlier:
> - bei `SelectPetri` einen Hovering-Hinweis neu ausgeben **sofern `petriSvgHighlighting = True` gesetzt wäre**, aber per Konfigurations-Checker zunächst erzwingen, dass immer `petriSvgHighlighting = False` gesetzt ist

src/Modelling/PetriNet/Reach/Draw.hs
```diff
@@ -74,7 +75,7 @@ isPetriDrawable
   -> m Bool
 isPetriDrawable petri cmd =
   let canDraw withoutPlaceNames = isNetDrawable (toPetriLike show show petri)
-        $ reachDrawSettings withoutPlaceNames cmd
+        $ reachDrawSettings withoutPlaceNames True cmd
   in canDraw True &&^ canDraw False
```
jvoigtlaender, line 80: Hmm, maybe actually this now shouldn't just test it with `True` for highlighting, but actually check the drawability in the way the things will actually be drawn. That is, without highlighting when there are no place names, and with highlighting when there are place names.

jvoigtlaender: Or actually, these things are independent and potentially four different versions need to be checked for drawability? Unless we are certain that the new parameter can impossibly have any impact on the drawability.

src/Modelling/PetriNet/Diagram.hs
```diff
@@ -105,6 +107,9 @@ cacheNet path pl drawSettings@DrawSettings {..} =
       ++ short withSvgHighlighting
       ++ short withGraphvizCommand
       ++ ".svg"
+    disableHover = if withSvgHighlighting
```
jvoigtlaender: This should have a different name, since it does not *always* disable the hovering.

jvoigtlaender:
For `SelectPetri` and `MatchPetri`, there seems to be no code in this PR so far that would actually disable the hovering?

src/Modelling/ActivityDiagram/MatchPetri.hs
jvoigtlaender: I was a bit confused here. This `Bool` does not determine whether the hovering comment is displayed or not, only in which form it is.

But the real point here and below is that the option from the config should influence whether or not the hovering comment is displayed at all (and whether hovering is indeed applied in the diagram).

jvoigtlaender:
It seems no such extra code (my last comment) is needed, because the relevant option is already having its effect?

However, for `SelectPetri` it would probably be better to always turn the hovering on in the Petri net displayed as part of the feedback.

jvoigtlaender:
Another problem is that when hovering is turned on for `MatchPetri`, say, then the wording of the `hoveringInformation` text now actually promises too much. It says that "respective diagram elements" will be highlighted. However, nothing at all will be highlighted in the displayed activity diagrams.

src/Modelling/PetriNet/Reach/Draw.hs
nimec01: The only thing that the `highlightSvg` parameter influences is whether a CSS class is added to the SVG or not. That does not influence drawability. The label annotations are/were the problematic part.

nimec01:
> Another problem is that when hovering is turned on for `MatchPetri`, say, then the wording of the `hoveringInformation` text now actually promises too much. It says that "respective diagram elements" will be highlighted. However, nothing at all will be highlighted in the displayed activity diagrams.

[Assuming activity diagrams don't support highlighting]
But that would apply to all tasks where a `hoveringInformation` and activity diagram are present at the same time, right? Not specifically related to the changes here.

jvoigtlaender:
> [Assuming activity diagrams don't support highlighting]

Which so far is the case. These diagrams are created via a completely different route, not using the `diagrams` library at all (on which the current highlighting feature depends).

> But that would apply to all tasks where a `hoveringInformation` and activity diagram are present at the same time, right?

Yes, but before this PR here there were no such tasks at all.

jvoigtlaender:
The immediate solution seems to be to create a special version of `hoveringInformation` with revised wording and use that here.

jvoigtlaender:
> `diagrams` library [...] (on which the current highlighting feature depends)

Maybe that's not accurate.

PR #575, opened 2026-02-03
record task 46-51 configs

guard-who:


jvoigtlaender:
Do we know that all the instances generated from the configs are okay in terms of drawing the Petri nets using the prescribed renderer? It seems neither SelectPetri nor MatchPetri check for successful rendering at generation time.

guard-who:
> Do we know that all the instances generated from the configs are okay in terms of drawing the Petri nets using the prescribed renderer?

I have just checked them manually. Fortunately, all of them are rendered successfully.

> It seems neither SelectPetri nor MatchPetri check for successful rendering at generation time.

It seems so. Perhaps, it would be better to add drawability check for them to avoid potential rendering errors.

PR #591, opened 2026-02-09
Update super-linter version to 8.5.0

jvoigtlaender:


jvoigtlaender:
@copilot: Can you add an exception so that BIOME_FORMAT will not anymore complain about `.github/linters/.jscpd.json`?

jvoigtlaender:
@copilot:

Can you make `codespell` not complain about any files under `.github` anymore, and also not about "shouldBe"?

Issue #475, opened 2025-12-12
Option for displaying "a mapped diagram" in feedback of `DifferentNames`


jvoigtlaender:
There could be an option in the task configuration that lets us choose between the following three things:
1. Have the feedback as currently, with no graphical output, just listing the solution mapping `[(x,3),etc.]`
2. Additionally reprint the class diagram from the task description but now with its relationships renamed according to the solution mapping.
3. Instead additionally reprint the object diagram from the task description but now with its links renamed according to the solution mapping.

jvoigtlaender:
For good visual matching with the reprintings, something like #463 might be relevant here. That is, already preparing the diagrams displayed in the task description in such a way that their layout is more similar to that of the later printed versions in the feedback. 

However, here we maybe cannot simply use the suggestion from #463 and already print the „real names“ into the task description, just set transparently. Because those names are the solution information nd we don’t want students to discover them from the diagrams, e.g. by using some tool that makes the transparent color non-transparent (if such a tool is available to them). Or by misusing the hovering feature (assuming the hovering kicks in also on transparent text and they could figure out the text length of labels by trying out where exactly the hovering effect kicks in).

jvoigtlaender:
To elaborate on the last comment here, the situation can, for example be like this:

The task description displays this as the object diagram:

<img width="942" height="1436" alt="Image" src="https://github.com/user-attachments/assets/1cd8610c-7e9b-471f-b0e3-974e0362dc5e" />

In the feedback this object diagram should be reprinted but with 1.-5. being replaced by names like `isOwnedby`, `uses` etc. that appeared in the class diagram depicted in the task description.

Since those names are much longer than the number labels, the layout might turn out much differently. But addressing this, by somehow already taking those longer labels into account while printing the above object diagram, must make sure that the information about which of the longer names will eventually go to which of the numbered places is not leaked in the object diagram (file) displayed in the task description.

(And there is a dual situation where the class diagram experiences this switch from short names to long names, with the same consequences and constraints.)

PR #602, opened 2026-02-17
Fix layout agreement issues on diagram reprinting

### LINKED ISSUE #463: Layout agreement with "reprinted class diagram"
When a CD occurs without relationship names in the original task but is then reprinted with names as part of the feedback, then the name printing can change the layout considerably.

For example,

Image

vs.

Image

Maybe it is possible to make this match more closely by already printing the original class diagram "with labels" that just take up space roughly equivalent to the their hidden names (or actually simply with their names, but in white font, that should remain white even when the highlighting feature is used)?

nimec01:
Closes #463 

It turned out that only these changes are necessary to have the same diagram layout (found that out while implementing the approach mentioned in https://github.com/fmidue/modelling-tasks/issues/463#issuecomment-3607105568). 

Before 
-----
(used diagram is `fromClassDiagram $ (Map.! 1) $ diagrams defaultMatchCdOdInstance`)
before_with_names
before_without_names

After
-----
after_with_names
after_without_names

My guess is that this not only applies to this example.

### DIFF
--- a/src/Modelling/CdOd/Output.hs
+++ b/src/Modelling/CdOd/Output.hs
@@ -150,7 +150,7 @@ relationshipArrow CdDrawSettings {..} marking isThick =
             $ limits compositionPart
           ]
         ++ concat [maybeToList marking | isThick]
-        ++ [toLabel compositionName | printNames]
+        ++ [toLabel compositionName]
       Aggregation {..} -> [
           arrowFrom oDiamond,
           edgeEnds Back,
@@ -162,7 +162,7 @@ relationshipArrow CdDrawSettings {..} marking isThick =
             $ limits aggregationPart
           ]
         ++ concat [maybeToList marking | isThick]
-        ++ [toLabel aggregationName | printNames]
+        ++ [toLabel aggregationName]
       Association {..} -> associationArrow ++ [
           TailLabel $ multiplicity
             (associationOmittedDefaultMultiplicity omittedDefaults)
@@ -172,7 +172,7 @@ relationshipArrow CdDrawSettings {..} marking isThick =
             $ limits associationTo
           ]
         ++ concat [maybeToList marking | isThick]
-        ++ [toLabel associationName | printNames]
+        ++ [toLabel associationName]
     associationArrow
       | printNavigations = [arrowTo vee, ArrowSize 0.4]
       | otherwise        = [ArrowHead noArrow]


jvoigtlaender:
How is the association name printing now suppressed at all in the code? I don’t get how the last diagram depicted above arises with the code after the commit.

nimec01:
All diagrams are generated using this function call
```haskell
cacheCd ((Modelling.CdOd.MatchCdOd.cdDrawSettings defaultMatchCdOdInstance) {printNames = False}) mempty (fromClassDiagram $ (Map.! 1) $ diagrams defaultMatchCdOdInstance) "/tmp/"
```
with `printNames = True` and `printNames = False` respectively.

---
`relationshipArrow` is a transitive call of `cacheCd` and does not have any impact on whether the names are shown or not. That happens in `connectWithPath`  (another transitive call from `cacheCd`).

However, `relationshipArrow` conditionally excluding the label changes that diagram "envelope", leading to different layouts (that's what AI told me at least).

jvoigtlaender:
Okay, thanks for the explanation! I wasn't aware that those lines/attributes are not what puts the labels into the rendered diagram.

In anticipation of https://github.com/fmidue/modelling-tasks/issues/475#issuecomment-3649051761, though, are these attributes like `toLabel compositionName`, now not anymore conditioned on `printNames`, somehow recoverable from the image? For example, by inspecting the source code and finding them inside some `svg` tags or so? Then we would have to be careful that (for tasks where the labelling information is meant to be secret) we don't create leakage here.

nimec01:
> In anticipation of [#475 (comment)](https://github.com/fmidue/modelling-tasks/issues/475#issuecomment-3649051761), though, are these attributes like `toLabel compositionName`, now not anymore conditioned on `printNames`, somehow recoverable from the image? For example, by inspecting the source code and finding them inside some `svg` tags or so? Then we would have to be careful that (for tasks where the labelling information is meant to be secret) we don't create leakage here.

The only difference between the two SVGs is that the`path`s used to display the relationship names are completely missing (in the diagram without names).

jvoigtlaender:
> only difference between the two SVGs

Is this comparison between
- diagram being generated before this PR with `printNames = True`, and
- diagram being generated after this PR with `printNames = False`?

If so, then I don't think that's the relevant comparison for assessing the potential for leakage from the fact that `relationshipArrow` now creates an attribute containing print names even when `printNames = False`.

nimec01:
> Is this comparison between
> 
> * diagram being generated before this PR with `printNames = True`, and
> * diagram being generated after this PR with `printNames = False`?
> 
> If so, then I don't think that's the relevant comparison for assessing the potential for leakage from the fact that `relationshipArrow` now creates an attribute containing print names even when `printNames = False`.

It's the difference between the two diagrams (`printNames = False` and `printNames = True`) after the changes in this PR.

jvoigtlaender:
Well, that's the same comparison, right? Given that this PR has zero semantic impact when `printNames = True`.

jvoigtlaender:
For leakage, the question would be: In the diagram generated after this PR with `printNames = False`, does the fact that `relationshipArrow` behaves as if `printNames = True` mean that the diagram contains information about the names that was not present before this PR in the diagram generated with `printNames = False`?

The above comparison is not able to answer this question, I would say.

nimec01:
> For leakage, the question would be: In the diagram generated after this PR with `printNames = False`, does the fact that `relationshipArrow` behaves as if `printNames = True` mean that the diagram contains information about the names that was not present before this PR in the diagram generated with `printNames = False`?
> 
> The above comparison is not able to answer this question, I would say.

There is no difference (except layouting) between the two diagrams before and after (with `printNames = False`)

jvoigtlaender:
Also in the svg source code?

nimec01:
> Also in the svg source code?

That is what I inspected.

PR #612, opened 2026-03-22
adjust match-cd-od task specific texts for variants

### LINKED ISSUE #611: adjust the demonstration of specific task texts in `MatchCdOdInstance` for its variant
We need to modify or hide some task texts and hints to make the variant [https://github.com/fmidue/llm-task-wrappers/pull/144](https://github.com/fmidue/llm-task-wrappers/pull/144) more sense.

guard-who:
close #611

### DIFF
--- a/src/Modelling/CdOd/MatchCdOd.hs
+++ b/src/Modelling/CdOd/MatchCdOd.hs
@@ -431,9 +431,9 @@ inputHelpText hasGivenCds = [
     Translated $ translations $ do
       english [iii|
         expresses that among the offered choices exactly
-        the object diagrams a and b conform to #{entityNameEn} 1 and
+        the object diagrams a and b are instances of class diagram #{entityNameEn} 1 and
         that none of the offered object diagrams
-        conform to #{entityNameEn} 2.
+        are instances of class diagram #{entityNameEn} 2.
         |]
       german [iii|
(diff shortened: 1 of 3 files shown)


src/Modelling/CdOd/MatchCdOd.hs
```diff
@@ -397,46 +401,55 @@ defaultMatchCdOdTaskText diagramCount instanceCount =  [
...
         Zum Beispiel drückt#{" "}|],
     Code . uniform . show $ matchingShow matchCdOdInitial,
     Translated $ translations $ do
       english [iii|
         expresses that among the offered choices exactly
-        the object diagrams a and b are instances of class diagram 1 and
+        the object diagrams a and b conform to #{entityNameEn} 1 and
         that none of the offered object diagrams
-        are instances of class diagram 2.
+        conform to #{entityNameEn} 2.
```
jvoigtlaender: You changed the English wording here. The German wording is unchanged, though.

The original wording was deliberately talking of "conforming to" in one part of the description and of "being instances of" in another part of the description.

jvoigtlaender:
The new specials could take the `Bool` as a parameter, so that the two `True`s in
```haskell
  DirectionsAdvice -> directionsAdvice True
  SimplifiedInformation -> simplifiedInformation True
```
would become configurable.

That would also make sense for `DifferentNames`.

PR #613, opened 2026-03-23
 Add example solutions to stored instances for `Reach` and `Deadlock` 

### LINKED ISSUE #565: add example solutions to (default) instances for Reach.Reach and Reach.Deadlock
In all these places: 

https://github.com/fmidue/modelling-tasks/blob/ae29a30cbd042a3638ea8a093f7491a10963a31d/example/src/Modelling/PetriNet/PetriReach/Instance.hs#L49-L50

https://github.com/fmidue/modelling-tasks/blob/ae29a30cbd042a3638ea8a093f7491a10963a31d/src/Modelling/PetriNet/Reach/Deadlock.hs#L327-L328

https://github.com/fmidue/modelling-tasks/blob/ae29a30cbd042a3638ea8a093f7491a10963a31d/src/Modelling/PetriNet/Reach/Reach.hs#L624-L625

https://github.com/fmidue/modelling-tasks/blob/ae29a30cbd042a3638ea8a093f7491a10963a31d/example/src/Modelling/PetriNet/PetriDeadlock/Instance.hs#L42-L43

the first occurrence of `[]` should be replaced by a correct solution (and `maxDisplayedSolutions` be set to `1` where it isn't already).

enellis:
_Closes #565._

### DIFF
--- a/test/Modelling/PetriNet/Reach/ReachSpec.hs
+++ b/test/Modelling/PetriNet/Reach/ReachSpec.hs
@@ -437,8 +437,8 @@ spec = do
     it "passes reachEvaluation" $ do
       let sols = fromEither $ shortestSolutions defaultReachInstance
       results <- withTempDir $ \tempDir ->
-        forM sols $ \sol ->
-          runWithoutOutput $ reachEvaluation tempDir defaultReachInstance sol
+        forM sols $
+          runWithoutOutput . reachEvaluation tempDir defaultReachInstance
       results `shouldBe` (Just 1 <$ results)
 
 hasMinTransitionLength
@@ -458,4 +458,3 @@ hasMinTransitionLength p ts minL n =
           a <- S.toList ts,
           as <- transitionVariants (x-1)
           ]
-
--- a/test/Modelling/PetriNet/Reach/ReachSpec.hs
+++ b/test/Modelling/PetriNet/Reach/ReachSpec.hs
@@ -439,7 +439,7 @@ spec = do
       results <- withTempDir $ \tempDir ->
         forM sols $
           runWithoutOutput . reachEvaluation tempDir defaultReachInstance
-      results `shouldBe` (Just 1 <$ results)
+      results `shouldSatisfy` (all (== Just 1))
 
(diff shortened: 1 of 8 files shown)


jvoigtlaender:
For the `defaultReachInstance` and `defaultDeadlockInstance`, it would be nice to add a test each to the test suite, checking that each solution recorded in the instance indeed passes `reachEvaluation` and `deadlockEvaluation`, respectively.

To deal with the path/file IO during those tests, it should be possible to proceed analogously to here: https://github.com/fmidue/modelling-tasks/blob/b293f8795b2c80d888d96c1c6b9abf68a2fe83ca/test/Modelling/CdOd/DifferentNamesSpec.hs#L369-L372

jvoigtlaender:
@enellis, any progress here?

jvoigtlaender:
@horriblename, can you take it from here?

test/Modelling/PetriNet/Reach/ReachSpec.hs
```diff
@@ -426,6 +434,14 @@ spec = do
             }
       checkReachConfig config `shouldBe` Nothing
 
+  describe "defaultReachInstance" $ do
+    it "passes reachEvaluation" $ do
+      let sols = fromEither $ shortestSolutions defaultReachInstance
+      results <- withTempDir $ \tempDir ->
+        forM sols $ \sol ->
+          runWithoutOutput $ reachEvaluation tempDir defaultReachInstance sol
+      results `shouldBe` (Just 1 <$ results)
```
jvoigtlaender: What is this condition doing? I would have expected that `results` is a list of `Maybe`s now?

horriblename: yes it's a `NonEmpty (Maybe Rational)` which is the scores for of each solution

the right hand side is a list of "full marks" (`Just 1`)

jvoigtlaender: Then wouldn’t the line be better expressed with shouldSatisfy and without „recurrence“ to results?

PR #615, opened 2026-03-24
use proper ShowMapping options for DifferentNames task variants

### LINKED ISSUE #614: Modify configs of various `DifferentNames` tasks using correct `ShowMapping` options
refer to [https://github.com/fmidue/modelling-tasks/commit/b293f8795b2c80d888d96c1c6b9abf68a2fe83ca#commitcomment-179883382](https://github.com/fmidue/modelling-tasks/commit/b293f8795b2c80d888d96c1c6b9abf68a2fe83ca#commitcomment-179883382)

guard-who:
close #614

### DIFF
--- a/example/src/Modelling/CdOd/DifferentNames/Config.hs
+++ b/example/src/Modelling/CdOd/DifferentNames/Config.hs
@@ -350,7 +350,8 @@ task2025_15 :: DifferentNamesConfig
 task2025_15 = task2025_14 {
   objectProperties = (objectProperties task2025_14) {
     anonymousObjectProportion = 1 % 1
-    }
+    },
+  printSolution = ShowMappingAndReprintOD
   }
 
 {-|
@@ -360,7 +361,9 @@ share same instances as task2025_14
 share same concept injection as task2025_15
 -}
 task2025_16 :: DifferentNamesConfig
-task2025_16 = task2025_15
+task2025_16 = task2025_15 {
+  printSolution = ShowMappingAndReprintCD
+  }
 
 {-|
 points: 0.15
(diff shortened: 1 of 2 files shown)


guard-who:
For `task2025_21`, is it also meaningful to print `CD`? Because it only shows a scenario description instead of a `CD`, I think that printing `CD` may help students bridge the association between the description and the `OD`. In other words, should we add an option of printing both CD and OD?

jvoigtlaender:
Yes, probably in `task2025_21` both class and object diagram should be printed, with the injected concept names.

And maybe even for `task2025_22` one could argue that it would make sense to at least have the option of also printing the according class diagram in the feedback. In the sense of: "this could have been the underlying model".

A more general question arises here of whether even after these enhancements, the feedback texts for these two types of task would be fully appropriate. That is, whether the text printed by `differentNamesEvaluation` makes sense for these task variants.

Adapting those texts would probably only be possible by using `flex-tasks` configuration, since one really needs to "dynamically change the code", not just change stuff in the stored instances.

But this should probably be dealt with in a separate issue.

jvoigtlaender:
Within the issue here, maybe even for `task2025_13` and `task2025_14` a switch from `ShowMapping` to `ShowMappingAndReprintOD` would be reasonable, since even for the fully artificially named instances, it is probably easier for students (when trying to understand why the printed solution is the correct one) to compare a relabelled OD to the given OD then to mentally have to relabel the OD with the printed correct mapping on their own in order to check it against the original CD.

Probably the text in https://github.com/fmidue/modelling-tasks/blob/b293f8795b2c80d888d96c1c6b9abf68a2fe83ca/src/Modelling/CdOd/DifferentNames.hs#L641-L644 should be slightly adapted then, to not talk of "Compare ...".

PR #625, opened 2026-03-25
add rejectLongerThan for EnterAS

horriblename:
as I mentioned in https://git.uni-due.de/fmi/autotool-dev/-/work_items/76#note_250167


jvoigtlaender:
Your are missing updates in the `example` directory.

PR #628, opened 2026-03-25
store auxiliary CD3 in MatchCdOdInstance, and more feedback output in that task

### LINKED ISSUE #626: `MatchCdOd` should store the involved but hidden CD3 in a separate field
refer to [https://github.com/fmidue/llm-task-wrappers/issues/146](https://github.com/fmidue/llm-task-wrappers/issues/146)

guard-who:
close #626

### DIFF
--- a/legacy-app/generate/Modelling/CdOd/Generate/MatchCdOd.hs
+++ b/legacy-app/generate/Modelling/CdOd/Generate/MatchCdOd.hs
@@ -59,7 +59,7 @@ getRandomTask
   :: (MonadAlloy m, MonadFail m, RandomGen g)
   => Int
   -> MatchCdOdConfig
-  -> RandT g m (Map Int Cd, Map Char ([Int], AlloyInstance))
+  -> RandT g m (Map Int Cd, Cd, Map Char ([Int], AlloyInstance))
 getRandomTask searchSpace config = do
   (cd1, cd2, cd3, numClasses) <- getRandomCDs searchSpace config
   alloyInstances <- lift $ getODInstances config cd1 cd2 cd3 numClasses
@@ -67,7 +67,7 @@ getRandomTask searchSpace config = do
   case maybeRandomInstances of
     Nothing      -> getRandomTask searchSpace config
     Just randomInstances -> pure
-      (M.fromList [(1, cd1), (2, cd2)], M.fromList $ zip ['a' ..] randomInstances)
+      (M.fromList [(1, cd1), (2, cd2)], cd3, M.fromList $ zip ['a' ..] randomInstances)
 
 getRandomCDs
   :: (MonadFail m, RandomGen g)
--- a/src/Modelling/CdOd/MatchCdOd.hs
+++ b/src/Modelling/CdOd/MatchCdOd.hs
@@ -504,6 +504,16 @@ matchCdOdEvaluation path task@MatchCdOdInstance {..} sub' = do
       sol = fst <$> instances
       matching = toMatching (M.keys diagrams) sol
       refOnlyLetters = M.keys $ M.filter null sol
+      (refOnlyMidEn, refOnlyMidDe) =
+        if Special GivenCds `elem` taskText
+        then
+          ( "conform to none of the given class diagrams"
+          , "passen zu keinem der gegebenen Klassendiagramme"
+          )
+        else
+          ( "do not correspond to the given scenario description"
+          , "entsprechen der gegebenen Szenariobeschreibung nicht"
+          )
       what = translations $ do
         english "instances"
         german "Instanzen"
@@ -521,18 +531,18 @@ matchCdOdEvaluation path task@MatchCdOdInstance {..} sub' = do
           | otherwise -> do
               paragraph $ translate $ do
                 english [iii|
-                  None of the class diagrams shown above applies to the following object diagram(s):
+                  The following object diagram(s) #{refOnlyMidEn} in the reference solution:
                   |]
                 german [iii|
-                  Zu den folgenden Objektdiagrammen passt keines der oben gezeigten Klassendiagramme:
+                  Die folgenden Objektdiagramme #{refOnlyMidDe} (Referenzlösung):
                   |]
               code $ "[" ++ intercalate ", " (map (:[]) $ sort refOnlyLetters) ++ "]"
               paragraph $ translate $ do
                 english [iii|
-                  Consider the following reference class diagram conforming to them:
+                  Below is a reference class diagram that these object diagram(s) could conform to:
                   |]
                 german [iii|
-                  Betrachten Sie das folgende Referenz-Klassendiagramm, das zu ihnen passt:
+                  Nachfolgend ein Referenz-Klassendiagramm, zu dem diese Objektdiagramme passen können:
                   |]
               image $=<< cacheCd cdDrawSettings mempty Nothing (fromClassDiagram cd) path
               pure ()
--- a/src/Modelling/CdOd/MatchCdOd.hs
+++ b/src/Modelling/CdOd/MatchCdOd.hs
@@ -29,7 +29,7 @@ module Modelling.CdOd.MatchCdOd (
 
 import qualified Modelling.CdOd.CdAndChanges.Transform as Changes (transform)
 
-import qualified Data.Bimap                       as BM (fromList, insert, member)
+import qualified Data.Bimap                       as BM (fromList, insert, keys)
 import qualified Data.Map                         as M (
   adjust,
   elems,
@@ -174,7 +174,7 @@ import Data.Bifunctor                   (Bifunctor (second))
 import Data.Bitraversable               (bimapM)
 import Data.Containers.ListUtils        (nubOrd)
 import Data.GraphViz                    (DirType (Forward))
-import Data.List                        (intercalate, singleton, sort)
+import Data.List                        ((\\), intercalate, singleton, sort)
 import Data.Map                         (Map)
 import Data.Maybe                       (fromJust, isJust, listToMaybe, mapMaybe, fromMaybe)
 import Data.Ratio                       ((%))
@@ -886,8 +886,7 @@ renameInstance inst@MatchCdOdInstance {..} names' nonInheritances' = do
       bmNames  = BM.fromList $ zip names names'
       bmNonInheritances = BM.fromList $ zip nonInheritances nonInheritances'
       bmWithIdForUnmappedKeys bm ks =
-        foldr (\k -> BM.insert k k) bm
-          [k | k <- nubOrd ks, not (k `BM.member` bm)]
+        foldr (\k -> BM.insert k k) bm (ks \\ BM.keys bm)
       bmNamesForReferenceCd =
         bmWithIdForUnmappedKeys bmNames (maybe [] classNames hiddenReferenceCd)
       bmNonInheritancesForReferenceCd =
--- a/src/Modelling/CdOd/MatchCdOd.hs
+++ b/src/Modelling/CdOd/MatchCdOd.hs
@@ -176,7 +176,7 @@ import Data.Containers.ListUtils        (nubOrd)
 import Data.GraphViz                    (DirType (Forward))
 import Data.List                        ((\\), intercalate, singleton, sort)
 import Data.Map                         (Map)
-import Data.Maybe                       (fromJust, fromMaybe, isJust, listToMaybe, mapMaybe)
+import Data.Maybe                       (fromJust, isJust, listToMaybe, mapMaybe, fromMaybe)
 import Data.Ratio                       ((%))
 import Data.String.Interpolate          (iii)
 import GHC.Generics                     (Generic)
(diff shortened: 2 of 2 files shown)


legacy-app/generate/Modelling/CdOd/Generate/MatchCdOd.hs
```diff
@@ -59,15 +59,15 @@ getRandomTask
...
+  -> RandT g m (Map Int Cd, Cd, Map Char ([Int], AlloyInstance))
 getRandomTask searchSpace config = do
   (cd1, cd2, cd3, numClasses) <- getRandomCDs searchSpace config
   alloyInstances <- lift $ getODInstances config cd1 cd2 cd3 numClasses
   maybeRandomInstances <- takeRandomInstances alloyInstances
   case maybeRandomInstances of
     Nothing      -> getRandomTask searchSpace config
     Just randomInstances -> pure
-      (M.fromList [(1, cd1), (2, cd2)], M.fromList $ zip ['a' ..] randomInstances)
+      (M.fromList [(1, cd1), (2, cd2)], cd3, M.fromList $ zip ['a' ..] randomInstances)
```
jvoigtlaender, line 70: Was the change here in the `legacy-app` forced?

guard-who: In [https://github.com/fmidue/modelling-tasks/pull/628/changes#diff-174f23dd6c13b301cd1d8d1902dcaaefb247b53412238911f302e72d158d908aL55](https://github.com/fmidue/modelling-tasks/pull/628/changes#diff-174f23dd6c13b301cd1d8d1902dcaaefb247b53412238911f302e72d158d908aL55), signature of `getMatchCdOdTask` was changed, so the change here was enforced.

src/Modelling/CdOd/MatchCdOd.hs
```diff
@@ -777,7 +777,9 @@ defaultMatchCdOdInstance = MatchCdOdInstance {
 classAndNonInheritanceNames :: MatchCdOdInstance -> ([String], [String])
 classAndNonInheritanceNames inst =
   let names = nubOrd $ concatMap classNames (diagrams inst)
+        ++ classNames (auxiliaryCd inst)
       nonInheritances = nubOrd $ concatMap associationNames (diagrams inst)
+        ++ associationNames (auxiliaryCd inst)
```
jvoigtlaender: Are we sure these changes will have no adverse effect on current usage of the task type? That is, currently the „hidden generated CD“ cannot influence what happens when renaming the displayed CDs for variability purposes. But afterwards it can. Does this change anything for how the task is „normally“ used.

jvoigtlaender:
If we will anyway prepare the „auxiliary CD“ to be preserved, we could directly make use of it in feedback of the „normal task“. Here is a suggestion:

Make that field a `Maybe` (and probably give it a different name in line with the proposal given here now). During generation of instances, it is filled with `Just`. But a hand-written instance, for example, could put `Nothing` there. If it is filled with `Just`, then during feedback generation, if there were ODs that didn‘t fit to CD1 and also not to CD2, it is printed out to students, something like: „The object diagrams so-and-so, while not fitting to CD1 or CD2, fits to the following class diagram: … the previously hidden one.“

src/Modelling/CdOd/MatchCdOd.hs
guard-who: It seems we can't guarantee the desired effect. If the auxiliary CD contains names not shown on CD1 and CD2, renaming might result in some of five ODs including names not displayed on the auxiliary CD, which could serve as clues for students when choosing answers.

guard-who:
> Make that field a `Maybe` (and probably give it a different name in line with the proposal given here now). During generation of instances, it is filled with `Just`. But a hand-written instance, for example, could put `Nothing` there. If it is filled with `Just`, then during feedback generation, if there were ODs that didn‘t fit to CD1 and also not to CD2, it is printed out to students, something like: „The object diagrams so-and-so, while not fitting to CD1 or CD2, fits to the following class diagram: … the previously hidden one.“

While this proposal makes the storage of the auxiliary CD optional and allows students to finally understand why the displayed ODs show names that are not shown, will it still serve as a clue for students to select answers?

Could we keep the original renaming pool (CD1 and CD2) and rename CD3's names that appeared on CD1 and CD2, while keeping the new names as they are? Of course, this would require some changes on the `llm-task-wrappers` side, mainly considering the proper preparation of the auxiliary CD when preparing for the Prompt.

src/Modelling/CdOd/MatchCdOd.hs
jvoigtlaender: I don’t understand what you are saying here.

guard-who: I misunderstood before. I overlooked the fact that during renaming, `bimap` should ensure the consistency of names displayed globally in CDs and ODs. Therefore, it's impossible for a name that exists in CD3 but not in CD1 or CD2 to appear in all five ODs. So, your concern above is that adding the element names in CD3 to the renaming pool will change the distribution of variability and affect the original („normally“ used) task (we are not sure if this will have any potential side effects).

So should we, based on your suggestion below, also ensure that the names in CD3 doesn't enter the renaming pool, and when renaming CD3, to prevent errors caused by names that don't exist in CD1 and C2, reconstruct the `bimap` for it, renaming the CD3's element names that exist in CD1 and CD2 according to the rename pool, but keeping the new names unchanged?

jvoigtlaender: Well, you could use one and the same mapping for all renamings, but make sure that this mapping is the identity on all names-that-appear-only-in-CD3.

src/Modelling/CdOd/MatchCdOd.hs
```diff
@@ -126,7 +127,7 @@ import Modelling.Types (
   showLetters,
   )
 
-import Control.Applicative              (Alternative ((<|>)))
+import Control.Applicative              (Alternative, (<|>))
```
jvoigtlaender: Why this change?

src/Modelling/CdOd/MatchCdOd.hs
```diff
@@ -482,23 +487,55 @@ matchCdOdSyntax task sub = addPretext $ do
...
         german "Instanzen"
       solution =
-        if showSolution task
+        if showSolution
         then Just . (DefiniteArticle,) . show . matchingShow
           $ matchCdOdSolution task
         else Nothing
-  multipleChoice what solution matching sub
+  reRefuse (multipleChoice what solution matching sub) $ do
+    case hiddenReferenceCd of
```
jvoigtlaender: Won't you be displaying all of the below now even in case that `showSolution` was set to `False`?

src/Modelling/CdOd/MatchCdOd.hs
```diff
@@ -482,23 +487,55 @@ matchCdOdSyntax task sub = addPretext $ do
...
+      Nothing -> pure ()
+      Just cd
+        | null refOnlyLetters -> pure ()
+        | otherwise -> do
+            paragraph $ translate $ do
+              english [iii|
+                None of the class diagrams shown above applies to the following object diagram(s):
```
jvoigtlaender: If `showSolution` was indeed set to `True`, the students do already know that those particular object diagrams do not conform to any of the previously given class diagrams, so the wording here is weird, as if telling the students something new.

Also, the language/use of concepts in the wording is weird. Nowhere else do we speak of a class diagram "applying" to an object diagram, or (below) of "a class diagram conforming to object diagrams".

Other aspects for the wording around here:
- If there is only one CD in the task instance, will the wording make sense?
- If a scenario description instead of a CD is displayed in the task description, will the wording make sense?

src/Modelling/CdOd/MatchCdOd.hs
```diff
@@ -802,13 +885,24 @@ renameInstance inst@MatchCdOdInstance {..} names' nonInheritances' = do
   let (names, nonInheritances) = classAndNonInheritanceNames inst
       bmNames  = BM.fromList $ zip names names'
       bmNonInheritances = BM.fromList $ zip nonInheritances nonInheritances'
+      bmWithIdForUnmappedKeys bm ks =
+        foldr (\k -> BM.insert k k) bm
+          [k | k <- nubOrd ks, not (k `BM.member` bm)]
```
jvoigtlaender: Why is the `nubOrd` needed? Can `classNames` or `associationNames` ever contain duplicates?

Instead of `not ... member`, use `notMember`.

And possibly it would even make more sense to use list difference `\\`?

src/Modelling/CdOd/MatchCdOd.hs
```diff
@@ -802,13 +885,24 @@ renameInstance inst@MatchCdOdInstance {..} names' nonInheritances' = do
...
+      bmNamesForReferenceCd =
+        bmWithIdForUnmappedKeys bmNames (maybe [] classNames hiddenReferenceCd)
+      bmNonInheritancesForReferenceCd =
+        bmWithIdForUnmappedKeys bmNonInheritances (maybe [] associationNames hiddenReferenceCd)
```
jvoigtlaender: Why the above `maybe []` calls? Is that really necessary? What about using `fromJust` here? These maps will never be used when the `hiddenReferenceCd` was `Nothing`.

src/Modelling/CdOd/MatchCdOd.hs
```diff
@@ -886,8 +886,7 @@ renameInstance inst@MatchCdOdInstance {..} names' nonInheritances' = do
       bmNames  = BM.fromList $ zip names names'
       bmNonInheritances = BM.fromList $ zip nonInheritances nonInheritances'
       bmWithIdForUnmappedKeys bm ks =
-        foldr (\k -> BM.insert k k) bm
-          [k | k <- nubOrd ks, not (k `BM.member` bm)]
+        foldr (\k -> BM.insert k k) bm (ks \\ BM.keys bm)
```
jvoigtlaender: Arguably, shouldn’t even have to call keys here, since in each case of the bm, you already know what the keys are, as you have just created the bms yourself a few lines above.

src/Modelling/CdOd/MatchCdOd.hs
guard-who: `mapM` does prevent `hiddenReferenceCd = Nothing` from being processed, but are there any side effects to using `maybe` here? Is it because the type is cleaner? I just feel that deriving `maybe` solely from the type might be safer?

jvoigtlaender: When I see „maybe []“ in code, I assume that this „[]“ will at some point be used. But that‘s not true here. It’s dead code. Using fromJust will document that fact. Also, ghc-hpc would flag that [] expression.

src/Modelling/CdOd/MatchCdOd.hs
```diff
@@ -504,6 +504,16 @@ matchCdOdEvaluation path task@MatchCdOdInstance {..} sub' = do
       sol = fst <$> instances
       matching = toMatching (M.keys diagrams) sol
       refOnlyLetters = M.keys $ M.filter null sol
+      (refOnlyMidEn, refOnlyMidDe) =
+        if Special GivenCds `elem` taskText
+        then
+          ( "conform to none of the given class diagrams"
```
jvoigtlaender: It could also be only one class diagram in an instance. See the commit I pointed you to a few days ago.

jvoigtlaender:
I will finish the wording myself.

jvoigtlaender:
Before merging this, probably a bit further work is needed to adapt to the "scenario case".

In particular, before the `case hiddenReferenceCd of` line, shouldn't a scenario-based version of the task output the class diagram underlying the verbally given scenario?

guard-who:
> Before merging this, probably a bit further work is needed to adapt to the "scenario case".
> 
> In particular, before the `case hiddenReferenceCd of` line, shouldn't a scenario-based version of the task output the class diagram underlying the verbally given scenario?

I also consider this, but should we create a new issue and implement this in a separate PR?

jvoigtlaender:
It’s just a few lines that also fit into the places anyway being touched here, right?

src/Modelling/CdOd/MatchCdOd.hs
```diff
@@ -482,23 +487,64 @@ matchCdOdSyntax task sub = addPretext $ do
...
+          english [iii|
+            Regarding the scenario description, the following class diagram would have been appropriate:
+            |]
+          german [iii|
+            Bezüglich der Szenariobeschreibung wäre das folgende Klassendiagramm geeignet gewesen:
+            |]
+        maybe
+          (error "There should not be no class diagram corresponding to the scenario description.")
+          (\cd -> image $=<< cacheCd cdDrawSettings mempty Nothing (fromClassDiagram cd) path)
+          (M.lookup 1 diagrams)
```
jvoigtlaender: If you are going to safe-guard against unexpected form of the `diagrams` field here (at the moment, rejecting when there is no `1` key in it), then I would argue for fully checking that it is as expected in a scenario-based variant. Namely, checking that it then contains only a single CD in `diagrams`. Something like:
```haskell
case M.toList diagrams of
  [(1, cd)] -> image ...
  _ -> error ... 
```

PR #635, opened 2026-03-31
add an option allowCommonSenseSolution in MatchCdOdConfig to support …

### LINKED ISSUE #632: Provide an option to control whether the LLM should take the `hiddenReferenceCd` into account when generating concepts
refer to [https://github.com/fmidue/llm-task-wrappers/issues/146#issuecomment-41255094990](https://github.com/fmidue/llm-task-wrappers/issues/146#issuecomment-4125509499)

guard-who:
close #632

### DIFF
--- a/app/match-cd-od.hs
+++ b/app/match-cd-od.hs
@@ -58,6 +58,7 @@ main = do
             maxNoCd = 2
             },
           omittedDefaultMultiplicities = defaultOmittedDefaultMultiplicities,
+          allowCommonSenseSolution = Nothing,
           printSolution    = False,
           timeout          = Nothing,
           withNonTrivialInheritance = Nothing,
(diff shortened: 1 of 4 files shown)


jvoigtlaender:
This would be the first time where a config option is introduced in `modelling-tasks` that has no meaning at all for users of "just" `modelling-tasks` without concept injection?

So maybe we should reconsider, and this additional configuration possibility should be introduced into the workflow of `llm-task-wrappers`. That is, a user would decide about this option when they start a concept injection run, not when they generate the original instances to be used as input for concept injection.

guard-who:
> This would be the first time where a config option is introduced in `modelling-tasks` that has no meaning at all for users of "just" `modelling-tasks` without concept injection?

This is also my concern, because this configuration option will not used here, I considered using the `Maybe` type (`Nothing` is used to represent no usage).

If this option is configured in `llm-task-wrappers`, I'm unsure if the `Maybe` type is still necessary. Because when it equals `Nothing`, it means we want to display CD3, but not force concepts to match CD3. This is clear for configuring formal constraints in Alloy, but for LLM, the impact of displaying CD3 without any constraints or instructions is unpredictable (it might affect finding good concepts for CD1 and CD2). Actually, we don't need `Nothing`, right?

jvoigtlaender:
Let‘s have that discussion (about Nothing being needed or not) in the other repo.

PR #647, opened 2026-06-01
provide reusable hint on the validity of class diagrams (to be used in the context of compositions)

### LINKED ISSUE #454: Hinweis auf Objektdiagrammbewohntheit
... wäre sinnvoll bei (manchen) `SelectValidCd` und gegebenenfalls verwandten Aufgaben zu geben.

Also in der Aufgabenstellung mitteilen, dass für Validität eines Klassendiagramms auch Objektdiagrammbewohntheit vorliegen soll/muss.

Vermutlich am ehesten/mindestens als `extraText` bei "besonders betroffenen" Fällen (nämlich wo "Doppeltenthaltensein per Komposition" relevant).

guard-who:
close #454

### DIFF
--- a/example/modelling-tasks-examples.cabal
+++ b/example/modelling-tasks-examples.cabal
@@ -17,6 +17,7 @@ library
       Modelling.ActivityDiagram.MatchPetri.Config
       Modelling.ActivityDiagram.SelectAs.Config
       Modelling.ActivityDiagram.SelectPetri.Config
+      Modelling.CdOd.Common
       Modelling.CdOd.DifferentNames.Config
       Modelling.CdOd.MatchCdOd.Config
       Modelling.CdOd.NameCdError.Config
--- a/example/src/Modelling/CdOd/Common.hs
+++ b/example/src/Modelling/CdOd/Common.hs
@@ -15,8 +15,8 @@ validClassDiagramAdviceTitle = M.fromList [
 
 validClassDiagramAdviceText :: M.Map Language String
 validClassDiagramAdviceText = M.fromList [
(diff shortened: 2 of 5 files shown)


jvoigtlaender:
Is this really only relevant for `SelectValidCd` tasks?

The issue said:
> `SelectValidCd` und gegebenenfalls verwandten Aufgaben

Maybe there are others where the hint could be relevant? Maybe some `RepairCd` tasks? Maybe some `NameCdError` tasks?

guard-who:
Here are some clarifications regarding this PR:
1. The validity advice is added only to task configs prefixed with `task_2025`.

2. The validity advice is added only to task configs containing `compositionCycle = True` or `wrongCompositionLimit = True`. The other cases are more akin to direct syntax errors within the class diagram itself, and thus do not require consideration of object diagrams. Of course, the current validity advice could simply be applied to all tasks if you think it better.

jvoigtlaender:
Actually, I think neither of the two cases mentioned under 2. should invoke the relevant advice. They are also both "just" syntactical violations directly at the class level:
- A cycle of compositions is syntactically wrong, even before object instantiation is considered.
- The `wrongCompositionLimit` setting is about cases where a composition whole has multiplicity `0..2` or `2` or something like that. Again, that's an immediatey syntax error that does not require considering object instances.

The hint was meant for cases where composition wholes are just using legal multiplicities like `1` and `0..1` but at object instantiation time it fails.

guard-who:
> The hint was meant for cases where composition wholes are just using legal multiplicities like `1` and `0..1` but at object instantiation time it fails.

I also feel that this is the most appropriate case. However, I haven't found any relevant configuration options in `allowedProperties` that explicitly enable this case. This is precisely the point I was puzzled about in #454 .

jvoigtlaender:
... Yes, it's possible that such cases are never generated, instead only `task2024_09` was a hand-crafted instances aimed at this semantic point.

PR #649, opened 2026-06-08
pretty print complex solution types

### LINKED ISSUE #580: Pretty-printing in "datatype answers"
For example, in `MatchAd` tasks, the "input format" looks like this:

Image

but then in feedback the correct solution is printed like this:

Image

horriblename:
fixes #580

> In general, one has to be careful, first checking: Is the student submission parsed according to ToDoc? (The point being that while read and show are inverses of each other, it it is not the case that render . toDoc and show always agree, not even modulo layout/whitespace.)

in the 3 tasks covered here, all use the deriving instance for `ToDoc` and `Reader`, and I've checked that `sol == runParser reader $ render $ toDoc sol`, so this isn't a problem


jvoigtlaender:
Wouldn't it be better to import `render` from `Autolib.ToDoc`?

Issue #637, opened 2026-04-17
possible "syntax checks" for `FindAuxiliaryPetriNodes`


jvoigtlaender:
- none of the counts given is negative
- the counts of auxiliary nodes is not greater than the sum of the non-auxiliary node counts (is that an invariant that really always holds?)
- none of the counts is greater than the number of activity diagram elements on the screen (is that something that can be made precise, and is safe to say?)

jvoigtlaender:
At the very least, the first point could be introduced now. And maybe also a bit consideration of the others.

horriblename:
> the counts of auxiliary nodes is not greater than the sum of the non-auxiliary node counts (is that an invariant that really always holds?)

for fork-join nodes with >= 4 branches, auxiliary nodes can exceed non-auxiliary ones like the following example (left is AD, right is Petrinet, orange nodes are auxiliary)

<img width="1200" height="580" alt="Image" src="https://github.com/user-attachments/assets/8a944328-cabc-4251-af8e-d458a3236ba9" />

The main cause is that, adding a branch to the fork-join part in the above method will only add 1 AD node, but add 2 auxiliary nodes, so the count of auxiliary nodes will exceed non-auxiliary nodes at some point (I think non-auxiliary count is always <= AD node count though, will have to double check if that is desired)

> none of the counts is greater than the number of activity diagram elements on the screen (is that something that can be made precise, and is safe to say?)

this is also violated in the same example, adding a branch adds a AD node but two auxiliary places, so count of auxiliary places will eventually exceed number of AD nodes

jvoigtlaender:
> I think non-auxiliary count is always <= AD node count though, will have to double check if that is desired

Yes, that would also be a useful sanity check. 

About the implementation, I think the checker should not take the instance as argument as long as it is not referencing its data (which might change with the above consideration).

jvoigtlaender:
closed by #654

PR #654, opened 2026-06-21
add syntax checks for FindAuxiliaryPetriNodes

horriblename:
#637

### DIFF
--- a/src/Modelling/ActivityDiagram/FindAuxiliaryPetriNodes.hs
+++ b/src/Modelling/ActivityDiagram/FindAuxiliaryPetriNodes.hs
@@ -276,8 +276,8 @@ findAuxiliaryPetriNodesSyntax fullCheck task sub = do
   pure ()
   where
     isNonNegative en de field = assertion (field sub >= 0) $ translate $ do
-      english [iii|Count of #{en} is at least zero?|]
-      german [iii|Anzahl der #{de} ist mindestens Null?|]
+      english [iii|Stated count of #{en} is not negative?|]
+      german [iii|Angegebene Anzahl von #{de} ist nicht negativ?|]
     adNodeCount = length $ Ad.nodes $ activityDiagram task
 
 findAuxiliaryPetriNodesEvaluation
--- a/src/Modelling/ActivityDiagram/FindAuxiliaryPetriNodes.hs
+++ b/src/Modelling/ActivityDiagram/FindAuxiliaryPetriNodes.hs
@@ -270,8 +270,8 @@ findAuxiliaryPetriNodesSyntax fullCheck task sub = do
     pure ()
 
   assertion (countOfNonAuxiliaryNodes sub <= adNodeCount) $ translate $ do
-    english [iii|Count of non-auxiliary nodes not greater than count of elements in the activity diagram?|]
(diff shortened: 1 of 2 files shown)


src/Modelling/ActivityDiagram/FindAuxiliaryPetriNodes.hs
```diff
@@ -253,6 +255,31 @@ findAuxiliaryPetriNodesInitial = FindAuxiliaryPetriNodesSolution {
...
 
+findAuxiliaryPetriNodesSyntax
+  :: OutputCapable m
+  => Bool
+  -- ^ Whether to do a full check. If False, skips negative checks.
+  -> FindAuxiliaryPetriNodesInstance
+  -> FindAuxiliaryPetriNodesSolution
+  -> LangM m
+findAuxiliaryPetriNodesSyntax fullCheck task sub = do
+  when fullCheck $ do
```
horriblename, line 266: This is to use the `min=0` attribute for `<input>` in the input form variant instead of a syntax check

src/Modelling/ActivityDiagram/FindAuxiliaryPetriNodes.hs
```diff
@@ -256,18 +257,28 @@ findAuxiliaryPetriNodesInitial = FindAuxiliaryPetriNodesSolution {
...
+    isNonNegative "non-auxiliary nodes" "Nicht-Hilfsknoten" countOfNonAuxiliaryNodes
+    isNonNegative "auxiliary places" "Hilfsstellen" countOfAuxiliaryPlaces
+    isNonNegative "auxiliary transitions" "Hilfstransitionen" countOfAuxiliaryTransitions
+    pure ()
+
+  assertion (countOfNonAuxiliaryNodes sub <= adNodeCount) $ translate $ do
+    english [iii|Count of non-auxiliary nodes is not more than count of activity diagram nodes?|]
+    german [iii|Die anzahl der Nicht-Hilfsknoten nicht größer als die Anzahl der Knoten im Aktivitätsdiagramm|]
```
jvoigtlaender: Is it the case that actually the number asked here will always be uniquely determined, and the max is actually also the exact correct answer? If so, maybe that should not be included as a syntax check after all, as it would permit the following "solution strategy":
- start submitting `0`
- then repeatedly increase by `1` and resubmit
- until the syntax check stops complaining.

jvoigtlaender: Or the opposite, starting from a big number and counting down.

jvoigtlaender: But maybe it's not always exact, due to the treatment of end nodes?

horriblename: > But maybe it's not always exact, due to the treatment of end nodes?

yes that is correct

src/Modelling/ActivityDiagram/FindAuxiliaryPetriNodes.hs
jvoigtlaender: Has this
> use the `min=0` attribute for `<input>` in the input form variant instead of a syntax check

actually been done?

